PR #2, opened 2019-05-28
Implemented Player#getEffectivePermissions on Bukkit and Nukkit

McRuben:


manuelkollus:
I would add null checks in that case, what do you want to do if the IPermissble class is null? 

Also, you are not using the latest code style currently used on CloudNet. In retrospect it doesn't look the same anymore.

Yours sincerely,
Manuel Kollus

cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/permission/IPermissionManagement.java
```diff
@@ -271,4 +271,29 @@ default boolean tryExtendedGroups(IPermissionGroup permissionGroup, String group
 
         return false;
     }
+
+    default Collection<Permission> getAllPermissions(IPermissible permissible) {
```
manuelkollus: Why no null checks

cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/permission/IPermissionManagement.java
```diff
@@ -271,4 +271,29 @@ default boolean tryExtendedGroups(IPermissionGroup permissionGroup, String group
...
+        return this.getAllPermissions(permissible, null);
+    }
+
+    default Collection<Permission> getAllPermissions(IPermissible permissible, String group) {
+        if (permissible == null)
+            return null;
```
manuelkollus: You can get an empty collection back.

cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/permission/IPermissionManagement.java
```diff
@@ -271,4 +271,29 @@ default boolean tryExtendedGroups(IPermissionGroup permissionGroup, String group
...
+
+        Collection<Permission> permissions = new ArrayList<>(permissible.getPermissions());
```
manuelkollus: You can underline an line break

realPanamo:
About the code style not matching: Brackets at the next line is what neither the maintainers nor the most of the participants use or want. So it's fine in my opinion.

realPanamo:
I agree with the null checks

cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/permission/IPermissionManagement.java
GiantTreeLP: Yeah, returning an empty collection is the only correct answer here.

McRuben: fixed

realPanamo:
Please use the Google code style brought in by #3

PR #22, opened 2019-06-04
added logPlayerConnections into config of bridge module

Meldiron:


cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/BridgeConfiguration.java
```diff
@@ -24,4 +24,6 @@
 
     private Map<String, String> messages;
 
+    private Boolean logPlayerConnections;
```
realPanamo: Please use a primitive here, no need for an object

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/node/listener/NodeCustomChannelMessageListener.java
```diff
@@ -45,22 +46,30 @@ public void handle(ChannelMessageReceiveEvent event)
             {
                 NetworkConnectionInfo networkConnectionInfo = event.getData().get("networkConnectionInfo", NetworkConnectionInfo.TYPE);
                 CloudNetDriver.getInstance().getEventManager().callEvent(new BridgeProxyPlayerLoginRequestEvent(networkConnectionInfo));
-                System.out.println(LanguageManager.getMessage("module-bridge-player-proxy-login-request")
-                    .replace("%uniqueId%", networkConnectionInfo.getUniqueId() + "")
-                    .replace("%name%", networkConnectionInfo.getName() + "")
-                    .replace("%proxy%", networkConnectionInfo.getNetworkService().getServerName() + "")
-                );
+
+                BridgeConfiguration bridgeConfiguration = CloudNetBridgeModule.getInstance().getBridgeConfiguration();
```
realPanamo: Please only take the Conifguration once and make it a class variable, in the constructor would be good

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/node/listener/NodeCustomChannelMessageListener.java
```diff
@@ -45,22 +46,30 @@ public void handle(ChannelMessageReceiveEvent event)
...
+                if(bridgeConfiguration.getLogPlayerConnections() == true) {
```
realPanamo: if(boolean == true) is redunant, use
if(boolean)

PR #47, opened 2019-07-13
NullPointerException Fix

ProxieTV:
Fixed the NullPointerException which was thrown when service not exists


ProxieTV:
Don't know why

PR #49, opened 2019-07-13
Now CloudNet takes care about custom properties

ProxieTV:
Added ability to define properties when creating ServiceInfoSnapshot and ServiceConfiguration.
Instead of overwrite those, now the Cloud cares about them.

Reached this by adding Properties as parameter to be able to use them on creation and storage of the Snapshot.

Feature could be useful to store easiely custom properties and query them later to do something with them.
This makes it easy to create PrivateServers and other servers which are needing useful information of their parent.

Tested with normal Document values and null values, it is working.


realPanamo:
This PR contains breaking api changes. You could add a method to give the service properties, but please dont do via the constructor or the create methods. Also there are commits which I already merged in here.

PR #52, opened 2019-07-13
New command

McRuben:
- [ ] breaking changes
- [x] no breaking changes

Added copy command to easily copy services to their template


realPanamo:
Please remove the ``created by ...`` in the comment, and let IntelliJ reformat everything, so it fits to the current style

realPanamo:
Please also don't make the whole template one argument, do it similar to the other commands, first argument the storage, second the prefix, last the name.

McRuben:
It is similar to other commands, just take a look at the create command

realPanamo:
Ah, I didn't see that the template is an optional argument

PR #57, opened 2019-07-14
BugFix, More ingame messages

McRuben:
- [ ] breaking changes
- [x] no breaking changes

- added missing import (from another pull request)
- added ingame messages when a server starts/stops


realPanamo:
Please dont call it "server", call it "service", because these are proxies too.

realPanamo:
I mainly meant the name "server" in the code, so the methods and also the name of the language-strings. Please rename that too.

PR #63, opened 2019-07-24
Create templates automatically if they don't exist

McRuben:
- [x] breaking changes (if anyone has created their own ITemplateStorage)
- [ ] no breaking changes

- added new method to the ITemplateStorage
- create the template in the command "tasks task/group <name> add template <storage> <prefix> <name>" if it doesn't exist


realPanamo:
Wouldn't normally merge this, but I thinks it's very unlikely, that someone created an own storage

PR #67, opened 2019-07-26
Bugfix

McRuben:
- [ ] breaking changes
- [x] no breaking changes

Removed throwing of an exception when the String given in the testStringParseToX method is not a number


realPanamo:
Please remove the ``value`` variable everywhere and return whether in the try and catch-block directly

PR #73, opened 2019-07-28
New option for a task

McRuben:
- [ ] breaking changes
- [x] no breaking changes

new option for the tasks (deletedFilesAfterStop)
all files included in this array will be deleted when a service of this task stops

### DIFF
--- a/cloudnet/src/main/java/de/dytanic/cloudnet/command/commands/CommandCreate.java
+++ b/cloudnet/src/main/java/de/dytanic/cloudnet/command/commands/CommandCreate.java
@@ -61,6 +61,7 @@ public final class CommandCreate extends CommandDefault implements ITabCompleter
                         serviceTask.getTemplates(),
                         serviceTask.getDeployments(),
                         serviceTask.getGroups(),
+                        serviceTask.getDeletedFilesAfterStop(),
                         serviceTask.getProcessConfiguration(),
                         serviceTask.getStartPort()
                 );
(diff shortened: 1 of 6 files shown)


cloudnet/src/main/java/de/dytanic/cloudnet/command/commands/CommandCreate.java
```diff
@@ -215,6 +218,10 @@ public void execute(ICommandSender sender, String command, String[] args, String
         } else
             deploy.addAll(deployments);
 
+        if (properties.containsKey("deletedFilesAfterStop")) {
+            deletedFilesAfterStop = Arrays.asList(properties.get("deletedFilesAfterStop").split(";"));
```
realPanamo: Please do it like it's done with the parameters "node" and "groups", in an one-liner.

cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
```diff
@@ -793,6 +794,16 @@ private int stop0(boolean force) {
             if (this.serviceConfiguration.getProcessConfig().getEnvironment().equals(ServiceEnvironmentType.BUNGEECORD)) {
                 new File(this.directory, "config.yml").delete();
             }
+            if (this.serviceConfiguration.getDeletedFilesAfterStop() != null) {
+                for (String s : this.serviceConfiguration.getDeletedFilesAfterStop()) {
```
realPanamo: Please don't call a variable s

PR #74, opened 2019-07-29
Option to disable removal of deployments

ProxieTV:
Since CloudNet clears the deployments into the service after deploying its ressources, in some cases it would be helpful having the ability to disable this.

An example usecase: Sometimes services have to be copied during runtime without stopping them, for example to copy data securely to prefent data loosing in crash cases.

Added overloaded methods for fully api-support

### DIFF
--- a/cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
+++ b/cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
@@ -502,7 +502,7 @@ final class JVMCloudService implements ICloudService {
     }
 
     @Override
-    public void deployResources() {
+    public void deployResources(Boolean removeDeployments) {
         for (ServiceDeployment deployment : this.deployments)
             if (deployment != null)
                 if (deployment.getTemplate() != null && deployment.getTemplate().getStorage() != null && deployment.getTemplate().getPrefix() != null &&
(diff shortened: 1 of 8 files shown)


realPanamo:
You can always copy services at runtime, and there was also a copy command added here #52

realPanamo:
I dont see an advantage when being able to disable this

cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
```diff
@@ -532,7 +532,7 @@ public void deployResources() {
                             deployment.getTemplate()
                     );
 
-                    this.deployments.remove(deployment);
+                    if (removeDeployments) this.deployments.remove(deployment);
```
realPanamo: Please don't use one-line if-statements

ProxieTV:
By using the copy command your software won't know when the copy operation is finished, so in some cases impossible to use without ugly delayed task (you don't know how long to wait, so to be sure that costs time)

So actually you have to add every deployment again after deploy ressources, that's no good solution.

realPanamo:
I just noticed the changes are only in the api, its fine then. But please correct your changes

cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
ProxieTV: i'm sorry, wrapper using comes from Kotlin development :D

ProxieTV:
yes i'll do that

ProxieTV:
done.

PR #77, opened 2019-07-30
Fixed MultiRoot problem

McRuben:
- [ ] breaking changes
- [x] no breaking changes

Use the ServiceInfoSnapshots for every node and not only from the node where the process is located at

### DIFF
--- a/cloudnet/src/main/java/de/dytanic/cloudnet/CloudNet.java
+++ b/cloudnet/src/main/java/de/dytanic/cloudnet/CloudNet.java
@@ -727,8 +727,10 @@ public final class CloudNet extends CloudNetDriver {
 
     @Override
     public ServiceInfoSnapshot getCloudServiceByName(String name) {
-        ICloudService service = Iterables.first(cloudServiceManager.getCloudServices().values(), cloudService -> cloudService.getServiceId().getName().equalsIgnoreCase(name));
-        return service != null ? service.getServiceInfoSnapshot() : null;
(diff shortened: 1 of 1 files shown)


cloudnet/src/main/java/de/dytanic/cloudnet/CloudNet.java
```diff
@@ -727,8 +727,10 @@ public void runCommand(ServiceInfoSnapshot serviceInfoSnapshot, String command)
 
     @Override
     public ServiceInfoSnapshot getCloudServiceByName(String name) {
-        ICloudService service = Iterables.first(cloudServiceManager.getCloudServices().values(), cloudService -> cloudService.getServiceId().getName().equalsIgnoreCase(name));
-        return service != null ? service.getServiceInfoSnapshot() : null;
+        return this.cloudServiceManager.getGlobalServiceInfoSnapshots().values().stream()
+                .filter(serviceInfoSnapshot -> serviceInfoSnapshot.getServiceId().getName().equalsIgnoreCase(name))
```
realPanamo: Comparing the uuid of the service might be better here

McRuben: What do you mean with that?

realPanamo: Services have uuids, use the name to compare them

realPanamo: Oh bruh

PR #82, opened 2019-08-03
Fixed classloader errors and warnings

ProxieTV:
Since CloudNet overwrote the scl field of SystemClassLoader, Java didn't like that.
Removing this part may occure some issues, because all aspects of using it by Dytanics mind are'nt known.
Tested shortly, most stuff seems to work.

Because the Jar File is already needed in JVMCloudService, it made sence to move the part from wrapper.
This made the ApplicationEnvironment useless, i'll suggest to remove it from the project. 
There are enough other ways to set the environment.

This have to be testet!! But i'm confident that these issues can be fixed quickly.
Please think about removal of ApplicationEnvironmentEvent --> Breaking Changes!


cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/runtime/RuntimeApplicationClassLoader.java
```diff
@@ -1,59 +0,0 @@
...
-
-    @Override
-    public boolean equals(Object obj) {
-        return super.equals(obj) || cloudNetWrapperClassLoader.equals(obj);
-    }
-
-    public ClassLoader getCloudNetWrapperClassLoader() {
-        return this.cloudNetWrapperClassLoader;
-    }
-}
```
realPanamo: Reason to remove this whole class?

cloudnet-launcher/src/main/java/de/dytanic/cloudnet/launcher/CloudNetLauncher.java
```diff
@@ -311,11 +312,11 @@ private static void startApplication(List<String> args, File targetDirectory, Co
         dependencyResources.add(target.toURI().toURL());
         dependencyResources.add(driverTarget.toURI().toURL());
 
-        ClassLoader classLoader = new RuntimeClassLoader(dependencyResources.toArray(new URL[0]), ClassLoader.getSystemClassLoader());
+        ClassLoader classLoader = new URLClassLoader(dependencyResources.toArray(new URL[0]));
+
         prepareApplication(variables);
 
         Method method = classLoader.loadClass(mainClazz).getMethod("main", String[].class);
-        method.setAccessible(true);
```
realPanamo: Why removing this?

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/sponge/SpongeCloudNetBridgePlugin.java
```diff
@@ -27,29 +27,6 @@
...
-
-
-            field = ClassLoader.class.getDeclaredField("scl");
-            field.setAccessible(true);
-            field.set(null, Thread.currentThread().getContextClassLoader());
-
-        } catch (Exception ex) {
-            ex.printStackTrace();
-        }
-
```
realPanamo: I dont think that this whole code was just unnecessary, removing it might cause problems with sponge.

cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/Wrapper.java
```diff
@@ -2279,75 +2275,27 @@ private boolean startApplication() throws Exception {
...
-        }
-
-        try {
-            return this.startApplication0(entry, environment);
-        } catch (Throwable throwable) {
-            throwable.printStackTrace(System.err);
+        if (entry.exists() && environment.isPresent()) {
+            return this.startApplication0(entry, environment.get());
+        } else {
+            return false;
```
realPanamo: How will the user know, that there's no environment the service can use? Please log an exception for this case.

realPanamo:
I'm unable to work on this or test this for the next two weeks, we could try to resolve the possible issues together and test everything when I'm back again.

cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/Wrapper.java
```diff
@@ -2279,75 +2275,27 @@ private boolean startApplication() throws Exception {
...
-
-        Field field = ClassLoader.class.getDeclaredField("scl");
-        field.setAccessible(true);
-        field.set(null, appClassLoader);
-
-        Class<?> main = Class.forName(mainClazz, true, ClassLoader.getSystemClassLoader());
+        ClassLoader classLoader = new URLClassLoader(new URL[]{file.toURI().toURL()}, Thread.currentThread().getContextClassLoader());
+        Class<?> main = Class.forName(mainClazz, true, classLoader);
         Method method = main.getMethod("main", String[].class);
-        method.setAccessible(true);
```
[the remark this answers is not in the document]
ProxieTV: main method have to be accessible by default, this is unnecessary

cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
```diff
@@ -601,19 +602,44 @@ private void startApplication() throws Exception {
...
+                        }
+                    }
+                    return false;
+                }).findFirst();
+
+        if (!environmentFile.isPresent()) {
+            long time = 10;
+            System.out.println(LanguageManager.getMessage("cloud-service-jar-file-not-found-error")
+                    .replace("%environment%", this.serviceConfiguration.getProcessConfig().getEnvironment().name()
+                    .replace("%time%", "test")));
```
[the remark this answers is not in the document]
ProxieTV: finished in latest commit

cloudnet-launcher/src/main/java/de/dytanic/cloudnet/launcher/CloudNetLauncher.java
ProxieTV: main method have to be accessible by default, this is unnecessary

cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/Wrapper.java
ProxieTV: already defined in JVMCloudService. Removing the ApplicationEnvironmentEvent will make this unnecessary, so please say your opinion about that

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/sponge/SpongeCloudNetBridgePlugin.java
ProxieTV: Well by adding the jar file to class path it make sence that this whole stuff won't be needed anymore. But i'll test it hardly this evening on some test sessions, wait for reply

cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/Wrapper.java
ProxieTV: also every errors which are thrown are walking into main method (with the "throws" in method call)

cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
```diff
@@ -601,19 +602,44 @@ private void startApplication() throws Exception {
         ));
 
         File wrapperFile = new File(System.getProperty("cloudnet.tempDir", "temp"), "caches/wrapper.jar");
+        Optional<File> environmentFile = Files.list(this.directory.toPath())
+                .map(Path::toFile)
+                .filter(file -> file.getName().endsWith(".jar"))
+                .filter(file -> {
+                    for (ServiceEnvironment environment : this.serviceConfiguration.getProcessConfig().getEnvironment().getEnvironments()) {
+                        if (file.getName().toLowerCase(Locale.ENGLISH)
```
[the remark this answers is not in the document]
ProxieTV: will be done in next commit

ProxieTV:
good point, if you want we can do. Also i'll take this evening to test everything i can expect and fix stuff if is needed, every results i'll post to you. Maybe also McRuben is interested in testing that. With the results you can decide what to do. In two weeks i'm on a jurney, too, but i think i'll take my laptop with me.

cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/runtime/RuntimeApplicationClassLoader.java
ProxieTV: everything this class has done is replaced with bether functions. Methods like getCloudNetWrapperClassLoader won't ever be call, this class is used nowhere. URLClassLoader makes every job. If this class is unnecessary and will never be needed (to use it the system class loader has to be touched -> we don't want that) it would make sence to delete it.

cloudnet/src/main/java/de/dytanic/cloudnet/service/JVMCloudService.java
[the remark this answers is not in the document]
realPanamo: Its still there

PR #84, opened 2019-08-04
Added methods to send channel messages to specific services

McRuben:
- [ ] breaking changes
- [x] no breaking changes

Added two new methods for sending a channel message over the network
 - first to send the message to a specific service
 - second to send the message to a specific task


realPanamo:
You tested every case?

McRuben:
Yes

PR #87, opened 2019-08-05
added velocity support for cloudnet module plugins

ProxieTV:
Added velocity-plugin.json to make it work with velocity.

Tested with actual latest build #167.

No breaking changes.


realPanamo:
Isn't the info via the annotation working anymore?

ProxieTV:
it was never working, because there's an check implemented for bungee.yml. Could be viewed in their sourcecode: https://github.com/VelocityPowered/Velocity/blob/4940f6b4471838e0ceab1942cd120c5b0046e489/proxy/src/main/java/com/velocitypowered/proxy/plugin/loader/java/JavaPluginLoader.java#L95

PR #89, opened 2019-08-05
Neutralization auf authors to CloudNetService

ProxieTV:
Considering futuristic changes and development on this system gets done by the CloudNetServiceTeam, neutralizing authors makes sence.
Main Effect which animated me for the changes: When Plugin gets enabled, now CloudNetService gets stated.
Whole Neutralization of CloudNet's authors is not finished with this PR, this is just a start with some nice effects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,15 +76,15 @@
 
 Linux / OSX
 ```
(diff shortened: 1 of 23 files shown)


realPanamo:
We wont be able to change the package names, sadly

PR #93, opened 2019-08-30
Fixed wrapper module config using wrong paths

McRuben:
- [x] breaking changes // maybe, if someone uses the wrapper modules and their configs (DriverModule#getConfig/saveConfig), which is very unlikely, because i've never seen anyone using this
- [ ] no breaking changes

Moved wrapper module configs from SERVICE/modules/MODULE/ to SERVICE/.wrapper/modules/MODULE


realPanamo:
Dont remove the tests, please

PR #95, opened 2019-08-30
Use random ports in tests

McRuben:
- [ ] breaking changes
- [x] no breaking changes

Generate a random port for each test for some networking software to prevent that the tests use ports that are already bound


cloudnet-driver/src/test/java/de/dytanic/cloudnet/driver/network/netty/NettyTestUtil.java
```diff
@@ -0,0 +1,34 @@
...
+        return generateRandomPort(10000, 50000);
+    }
+
+    public static boolean isPortAvailable(int port) {
+        try (ServerSocket serverSocket = new ServerSocket()) {
+            serverSocket.bind(new InetSocketAddress(port));
+            return true;
+        } catch (Exception e) {
+        }
+        return false;
```
realPanamo: Put the return false into the catch block and rename the exception to "ignored"

cloudnet-driver/src/test/java/de/dytanic/cloudnet/driver/network/netty/NettyTestUtil.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+import java.net.InetSocketAddress;
+import java.net.ServerSocket;
+import java.util.Random;
+
+public class NettyTestUtil {
+
+    public static int generateRandomPort(int min, int max) {
+        Random random = new Random();
```
realPanamo: Create a Random constant instead of creating one every time

cloudnet-driver/src/test/java/de/dytanic/cloudnet/driver/network/netty/NettyTestUtil.java
```diff
@@ -0,0 +1,34 @@
+package de.dytanic.cloudnet.driver.network.netty;
+/*
```
realPanamo: Remove the comment

PR #99, opened 2019-09-06
Added permission check for console commands on "/cloud"

McRuben:
- [x] breaking changes (breaks commands if someone uses the CommandInfo in a NodeModule which unlikely, but if this is the case, the developers of those commands should be able to fix this easily)
- [ ] no breaking changes

- Added some methods to get all console commands in the node from the wrapper
- Check for the permission when a player executes "/cloud" before the command is executed in the console


cloudnet-common/src/main/java/de/dytanic/cloudnet/common/concurrent/ITask.java
```diff
@@ -57,8 +57,6 @@ public void onCancelled(ITask<V> task) {
 
     Collection<ITaskListener<V>> getListeners();
 
-    Callable<V> getCallable();
```
realPanamo: Any reason to remove this? Would cause breaking changes.

cloudnet-modules/cloudnet-report/src/main/java/de/dytanic/cloudnet/ext/report/command/CommandReport.java
```diff
@@ -6,7 +6,7 @@
 import de.dytanic.cloudnet.CloudNet;
 import de.dytanic.cloudnet.cluster.IClusterNodeServer;
 import de.dytanic.cloudnet.command.Command;
-import de.dytanic.cloudnet.command.CommandInfo;
+import de.dytanic.cloudnet.common.command.CommandInfo;
```
realPanamo: huh. why this?

McRuben: The CommandInfo wouldn't be accessable from the driver

cloudnet-common/src/main/java/de/dytanic/cloudnet/common/concurrent/ITask.java
McRuben: Well, no, not really, forgot to remove this change. I will add this later

cloudnet-modules/cloudnet-report/src/main/java/de/dytanic/cloudnet/ext/report/command/CommandReport.java
realPanamo: huh, but isn't the common stuff in the driver?

McRuben: Yes, but the CommandInfo was in the node before and therefore not accessable from the driver

realPanamo: ah

Issue #1, opened 2019-05-25
Synchronize the Gson version with the one used in Bungeecord and Minecraft


GiantTreeLP:
The `cloudnet-common` module is completely compiled into the Bungeecord and Bukkit plugins.
Due to mismatching Gson versions, some usage of Gson in conjunction with CloudNet is broken and fails.

Example error message:

```
java.lang.LinkageError: loader constraint violation: when resolving method "de.dytanic.cloudnet.common.document.gson.JsonDocument.<init>(Lcom/google/gson/JsonObject;)V" the class loader (instance of net/md_5/bungee/api/plugin/PluginClassloader) of the current class, de/web/bungeeplugin/partysystem2/listener/Events, and the class loader (instance of sun/misc/Launcher$AppClassLoader) for the method's defining class, de/dytanic/cloudnet/common/document/gson/JsonDocument, have different Class objects for the type com/google/gson/JsonObject used in the signature
```

4zuree:
So... is this bug fixed? If not I'd like to take actions and fix it rn

PR #103, opened 2019-09-09
Updated Gson Version

ProxieTV:
- [ ] breaking changes
- [x] no breaking changes

Changed Gson version to newer one.
Done this to prevent warning of illegal reflective access to some constructor by gson.
Tested, seems to work. Fortunately the old version and new version of gson are not so much different.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ allprojects {
         //Dependencies
         dependencyLombokVersion = '1.18.6'
         dependencyH2Version = '1.4.197'
-        dependencyGsonVersion = '2.8.2'
+        dependencyGsonVersion = '2.8.5'
         dependencyNettyVersion = '4.1.36.Final'
         dependencyJLine2Version = '2.14.6'
         dependencyJAnsiVersion = '1.17.1'
--- a/build.gradle
+++ b/build.gradle
@@ -54,7 +54,7 @@ allprojects {
         //Dependencies
         dependencyLombokVersion = '1.18.6'
         dependencyH2Version = '1.4.197'
-        dependencyGsonVersion = '2.8.5'
+        dependencyGsonVersion = '2.8.0'
         dependencyNettyVersion = '4.1.36.Final'
         dependencyJLine2Version = '2.14.6'
         dependencyJAnsiVersion = '1.17.1'
--- a/cloudnet-common/src/main/java/de/dytanic/cloudnet/common/document/gson/JsonDocument.java
+++ b/cloudnet-common/src/main/java/de/dytanic/cloudnet/common/document/gson/JsonDocument.java
@@ -839,6 +839,6 @@ public class JsonDocument implements IDocument<JsonDocument> {
 
     @Override
     public Iterator<String> iterator() {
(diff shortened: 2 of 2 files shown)


realPanamo:
I think there was a conflict with bukkit/bungeecord which use the older version too. When changing the cloud's version, this caused errors in BungeeCord.

McRuben:
It already causes problems with the current version of Gson in the cloud, so this might fix this, but idk which version of Gson BungeeCord uses

ProxieTV:
It seems like Bungeecord is using version 2.8.0: https://github.com/SpigotMC/BungeeCord/blob/9e76966e0fc46e4bdc61cfe97f14cd8b98290d8c/chat/pom.xml#L25

Problems with compatability on gson only one user reported, he sayd that this causes when sending channel messages. In my case, channel messages are working and i don't get this error. 
I think the chance updating the version will cause more errors then before is lower than some thinks are got fixed in the latest version.

McRuben:
There isn't only the gson version of Bungee, Spigot's using much older versions

ProxieTV:
yeah, i think in most cases the systems are indepent of each other, so there only can be caused errors on situations when they interact with each other. 
But since the old version already weren't similar to the version of other systems, it won't cause more problems.

And thinking about that this cloud won't work with Java 14 without that, we must renew it at some time.

realPanamo:
The common module is compiled into bungee and spigot plugins. So it can cause issues. Just look at #1

realPanamo:
This is why the version should be kept in sync

ProxieTV:
yes, let's change the version to 2.8.0 (3 years old) temporally until some of us have motivation to change the process of compiling common into plugins. Because such dependents are uglyt.

ProxieTV:
Whats missing for merging this? With this the Gson Version is Syncronized with Bungeecord's gson version

realPanamo:
I'm actually not sure about this issue, the current used gson version in v3 works fine for me. What would you say @GiantTreeLP?

ProxieTV:
This should fix this issue until someone makes this cloud independent of other gson versions.
https://discordapp.com/channels/325362837184577536/536594898321670154/581604794263928852

GiantTreeLP:
The best solution would be, to move our dependencies (like gson) to a custom package. That way, they don't interfere with already present classes. Maven shade has the ability to do that and I think Gradle should be able to do this as well.

GiantTreeLP:
@ProxieTV and me solved this issue which also affected Netty and legacy versions of Spigot/Paper.

@ProxieTV I have assigned this to you. We have already discussed the need to include all dependencies in the artifacts. At least for the wrapper, this is a necessity in order to relocate the packages and go out of the way of the server software and plugins.

It would be beneficial to CloudNet to also include all of its dependencies (relocated or not) in its artifact, so that offline installation is possible and network disruptions don't matter.

GiantTreeLP:
Closes #1

ProxieTV:
Since in the newest gson version are still illegal reflective accesses, it won't change a lot. We have to wait for newer gson release, in which this will be fixed. 
So updating gron won't change a lot actually, so this could be closed.

PR #104, opened 2019-09-09
Changed some author fields

ProxieTV:
- [ ] breaking changes
- [x] no breaking changes

Changed some more author names.


realPanamo:
I think we should keep Dytanic in there, he made the whole base of the cloud usw. I would suggest "Dytanic & the CloudNet Community" or something like this

ProxieTV:
yeah, this sounds fine, i'll change this later

PR #109, opened 2019-09-13
Fixed some problems with task deletion

McRuben:
- [ ] breaking changes
- [x] no breaking changes

- delete files of tasks when they are deleted
- check if the ServiceTaskRemoveEvent is cancelled before deleting the task
- changed protected access to private access of some fields

### DIFF
--- a/cloudnet/src/main/java/de/dytanic/cloudnet/service/DefaultCloudServiceManagerConfiguration.java
+++ b/cloudnet/src/main/java/de/dytanic/cloudnet/service/DefaultCloudServiceManagerConfiguration.java
@@ -79,16 +79,16 @@ final class DefaultCloudServiceManagerConfiguration {
                     return FileVisitResult.CONTINUE;
                 }
             });
-        } catch (IOException e) {
-            e.printStackTrace();
+        } catch (IOException exception) {
+            exception.printStackTrace();
         }
     }
 
     public void save() {
         try {
             Files.createDirectories(TASKS_DIRECTORY);
-        } catch (IOException e) {
-            e.printStackTrace();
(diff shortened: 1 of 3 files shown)


cloudnet/src/main/java/de/dytanic/cloudnet/service/DefaultCloudServiceManagerConfiguration.java
```diff
@@ -95,6 +95,22 @@ public void save() {
             new JsonDocument(task).write(TASKS_DIRECTORY.resolve(task.getName() + ".json"));
         }
 
+        try {
+            Files.walkFileTree(TASKS_DIRECTORY, new SimpleFileVisitor<Path>() {
+                @Override
+                public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+                    String name = file.getFileName().toString();
+                    if (name.endsWith(".json") &&
+                            tasks.stream().noneMatch(serviceTask -> serviceTask.getName().equalsIgnoreCase(name.substring(0, name.length() - 5)))) {
```
realPanamo: Just check if name is equal to task name + .json

cloudnet/src/main/java/de/dytanic/cloudnet/service/DefaultCloudServiceManagerConfiguration.java
```diff
@@ -95,6 +95,22 @@ public void save() {
...
+                        Files.delete(file);
+                    }
+                    return FileVisitResult.CONTINUE;
+                }
+            });
+        } catch (IOException e) {
```
realPanamo: Naming :smart:

PR #119, opened 2019-09-18
Changed dependency handling into wrapper

ProxieTV:
- [ ] breaking changes
- [x] no breaking changes

Changed:
- Updated deprecated field settings in gradle file to conventional usuage
- Added shadow plugin to create shadowed jar files for wrapper
- Added shadow relocate functions to build config for inserting needed dependencies into wrapper jar file
- Removed Collection of all usable wrapper dependencies from classpath
- Added liberty to use our own dependency versions without relations to server software one's

Related Issues:
- Fixed 1.8 netty join problem
- Fixed compatibility issues that where caused by affection dependencies into classpath
- Fixed gson version issue (github): https://github.com/CloudNetService/CloudNet-v3/issues/1
- Fixed gson channelMessage issue (discord): https://discordapp.com/channels/325362837184577536/536594898321670154/581604794263928852
- Opened door to use this Pull Request with new gson Version: https://github.com/CloudNetService/CloudNet-v3/pull/103

Note: By this change the wrapper jar will grow to 4 MB, so also the launcher.jar file size will be affected. But this must be done, because dytanics solution with loading all dependencies into classpath will cause much more incompatability issues than it already does. If we really want to still keep the launcher size little, maybe downloading the wrapper jar from external server repo will be another solution. But the wrapper file has to be such big!

### DIFF
--- a/cloudnet-common/build.gradle
+++ b/cloudnet-common/build.gradle
@@ -1,5 +1,5 @@
 jar {
-    archiveName = cloudnetCommonFile
+    archiveFileName.set(cloudnetCommonFile)
 }
 
 dependencies {
--- a/cloudnet-driver/build.gradle
+++ b/cloudnet-driver/build.gradle
@@ -1,6 +1,6 @@
 jar {
 
-    archiveName = cloudnetDriverOutFileName
+    archiveFileName.set(cloudnetDriverOutFileName)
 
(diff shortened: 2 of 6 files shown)


GiantTreeLP:
Fixes ä1

PR #121, opened 2019-09-18
Ignore interrupt warning on server stop

ProxieTV:
- [ ] breaking changes
- [x] no breaking changes

Changes: 
- made the case of InterruptException ignored to prevent printing it
- since the thread have to be interrupted, this warning is the normal behaviour
- corrected little type fault :D

### DIFF
--- a/cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/Wrapper.java
+++ b/cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/Wrapper.java
@@ -2265,10 +2265,7 @@ public final class Wrapper extends CloudNetDriver {
                     if (diff < millis) {
                         try {
                             Thread.sleep(millis - diff);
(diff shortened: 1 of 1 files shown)


cloudnet-wrapper-jvm/src/main/java/de/dytanic/cloudnet/wrapper/Wrapper.java
```diff
@@ -2265,8 +2265,9 @@ private synchronized void start0() throws Exception {
                     if (diff < millis) {
                         try {
                             Thread.sleep(millis - diff);
-                        } catch (Exception exception) {
-                            exception.printStackTrace();
+                        } catch (InterruptedException ignored) {
+                        } catch (Exception e) {
```
realPanamo: This second catch clause is redundant

realPanamo:
Just ignoring an exception is not the best thing to do, I fixed the issue here: 4177a030b6db0cbf51c43136ef4bab905f0fb65e

Issue #25, opened 2019-06-06
Time to make some changes to CloudNet's codebase


ghost:
So, I "analyzed" CloudNet's code and oh boy, a lot of work needs to be done. By work I mean actual code and performance optimization, debugging etc.

Most things can probably be done using IntelliJ's advanced refactoring tools, like searching for something and replacing it, but there's also stuff on the list which needs to be done by hand. 

Pending:
- [x] *[approved]* Use builders where applicable
- [ ] *[approved]* Create data classes where applicable
- [ ] *[wip]* Replace pure string concats `abc += "def"` with StringBuilder appends 
- [x] *[approved]* Replace `java.io` classes with `java.nio` classes where applicable, might require `@Deprecated` declarations
- [x] *[approved]* Some abstract classes and interfaces are waayyy too fat. Requires separation. Might require `@Deprecated` declarations
- [ ] *[approved]* Return `Collection<?>` types instead of `List<?>`, `Set<?>` or any other extending interface or class of `Collection`. Might require `@Deprecated` declarations
- [ ] *[approved]* **Use lazy type initialization. (Let a field be initialized by a getter or setter, if the field is null. A lot of stuff in CloudNet might not be needed by anyone or anything)**

Work in progress:
- [x] Replace inline statements with block statements 
- [x] Calls to functions located in own class should use the `this.` accessor
- [x] Calls to functions located in super class should use the `super.` accessor **ONLY IF** the function is not overridden by the child class
- [x] Try catch blocks shouldn't silently dispose any exceptions/throwables. (Re-)Throw, print or log them
- [x] Parameter validations shouldn't silently dispose any exceptions. (Re-)Throw, print or log them
- [x] Re-arrange fields correctly
- [x] Re-arrange functions correctly
- [x] Separate chained field declarations
- [x] Validate the state of parameters (Is the int negative?, is the string blank even though it shouldn't be?, is the passed varargs array empty?)
- [x] Validate the parameter for null
- [x] Use Guava's Preconditions instead of own validation class

Done:
- [x] Lombok should not be used in such complex application. The only thing Lombok could be used for is generating `#equals()`, `#hashCode()` and `#toString()`. Using `@Getter`, `@Setter` and any other Lombok annotations is highly discouraged blurs a class' logic

*More to come*

byRoadrunner:
First of all thank you for this "analysis" of CloudNet, we´ll look into these issues and will do our best to fix them.
[...]

PR #186, opened 2019-12-07
Cleanup interfaces named starting with 'I'

CraftCoderr:
- [x] breaking changes
- [ ] no breaking changes

CloudNet have a relatively large code base and many interfaces named in C# style, which not the best practice for Java. This PR replaces interfaces named starting with `I`, with same named interfaces without `I`. I tried to make as less as possible breaking changes. 

### related issues/discussions
  
#25


CraftCoderr:
Another way is to use common style for naming. Well, it will be Java project with interfaces named in C#-style. It not good, not bad, it as is. But it better, than different styles within one big project.

realPanamo:
What exactly is breaking here? I can't accept breaking changes with class namings, sorry

realPanamo:
I'm not sure about this, reviewing this will take time and I might have to change some things.
Much afford with no benefits at all.

realPanamo:
I just checked - there are still a lot of classes left that are called with 'I' at the beginning

McRuben:
Some classes are completely removed instead of deprecating them

CraftCoderr:
> I just checked - there are still a lot of classes left that are called with 'I' at the beginning

Yes, work in progress, but i need to know if there some chances to merge, when work would be completed.



> Some classes are completely removed instead of deprecating them

In some cases interfaces duplication is useless, as in factory classes, which should return instances of new interfaces, but may have old interfaces in current implementation.

realPanamo:
Closing due to inactivity.

PR #226, opened 2020-02-04
Changed domain of sponsor

Kenuux:
- [ ] breaking changes
- [X] no breaking changes

Changed domain of sponsor

### related issues/discussions
  
No issues related to this


realPanamo:
EU-Hosting is a part of OpusX, so I think theres no need to change this ;)

byRoadrunner:
OpusX is the new name of EU-Hosting, so I think we should change it to the new name.

PR #227, opened 2020-02-09
Upgrades jline from 2 to 3

derklaro:
- [X] breaking changes
- [ ] no breaking changes

Upgrading jline version


derklaro:
Currently tests in progress related to all types of console animations... I'm not sure if they work as they should.

derklaro:
Tested the animations for download and setup. They work as they should. @derrop can you please recheck this?

derrop:
I will test everything later

PR #230, opened 2020-02-09
Add LabyMod module

derrop:
- [ ] breaking changes
- [x] no breaking changes

Adds a new module to the Cloud.
This module implements the LabyMod Discord RPC Protocol and shows what game mode a LabyMod user plays to their friends.


derklaro:
I think this is not a good module for the cloud system... It already has too much modules which are not required :(

realPanamo:
I think it's a cool feature, we will rework the default modules anyways, so we might remove this module from the defaul modules and just make it available via the ''module install'' command.

PR #238, opened 2020-02-28
WIP Introduce a new website for reports

einrobin:
- [ ] breaking changes
- [x] no breaking changes

Adds a website for reports to the report module as an alternative to the paste command.
image
It can be enabled/disabled by using the "webreport --local" command.

You can also upload a report to a server instead of hosting it on the internal web server using the "webreport" command


derklaro:
I think it's better to make this pull request a draft pull request

PR #243, opened 2020-03-13
Prevent heap space errors when coping a template from one storage to another

derklaro:
- [ ] breaking changes
- [X] no breaking changes

Prevents heap space overload by coping templates with java streams


derklaro:
I wasn't able to test the template storages ftp/sftp and the queued ftp template storage. Can somebody take a look for me?

derrop:
The SFTP storage almost works. The SFTP server is on the right side and the local storage on the left side. I copied it using the following command: "t copy ftp:Test/test local:Test/test"
image

juliarn:
FTP seems to work completely :peepoHappy:

derrop:
Between linux and windows? I think the problem with the SFTP is the same as the "clu push lt" between one node on linux and another node on windows.

juliarn:
Yes this could be the problem

derklaro:
@juliarn has you fixed the sftp storage or should i take a look?

juliarn:
I only tested FTP

derrop:
I think it should be easily fixable like I did it here https://github.com/CloudNetService/CloudNet-v3/commit/f4f4fc099f6655e5673f9209c9585e7c52d21822

derrop:
Copying to an SFTP Server works now.

derklaro:
Ah nice

derrop:
So I think we can merge this

PR #294, opened 2020-07-07
Use the forms of politeness in every translation

0utplay:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Use the forms of politeness in every translation and do not switch between

No difference


derklaro:
It looks like there are a lot of translations still not in the politeness form...

PR #328, opened 2020-09-23
Add license

juliarn:
Adds a license to the repository and to all downloadable Files of CloudNet.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -30,6 +30,9 @@ jobs:
           echo 'Creating CloudNet.zip file...';
           mkdir -p temp/;
           cp -r .template/* temp/;
+          
+          cp LICENSE temp/LICENSE;
+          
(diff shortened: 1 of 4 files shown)


juliarn:
Maybe we should create the CloudNet.zip via Gradle so we don't have to update the build files for Jenkins and Actions everytime.

byRoadrunner:
> 
> 
> Maybe we should create the CloudNet.zip via Gradle so we don't have to update the build files for Jenkins and Actions everytime.

Why should we do that? Is this a huge problem?

juliarn:
> > 
> > 
> > Maybe we should create the CloudNet.zip via Gradle so we don't have to update the build files for Jenkins and Actions everytime.
> 
> Why should we do that? Is this a huge problem? I dont see any advantage in moving it.

Just using the advantages we have through Gradle. By doing it with Gradle we don't have to do it twice in the build files.

PR #370, opened 2020-10-15
Set the tab header & footer to null if there is no configuration

### LINKED ISSUE #365: The tablist is set also if the tabListConfigurations array is empty
**Description of the bug:**

The tabListConfigurations array is empty. The tablist is still not like in vanilla. 

**Steps to reproduce this bug:**

Clear the tabListConfigurations array in the SyncProxy module. Reload the module.

**Expected behavior:**

The tablist should be completely in vanilla style

**Screenshots:**

https://imgur.com/a/iUYBjjd

**Environment and used hardware:**
 OS: Win 10
 Version:  3.4.0-SNAPSHOT-530ab40

0utplay:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

The tablist header & footer are now set to null instead of an empty string

https://imgur.com/a/5IeGqL3
Due to this change the tablist is now like in vanilla

Fixes #365


juliarn:
The root cause is that there wasn't a check for the index before setting the tab for a player, I fixed this in the above commit

PR #385, opened 2020-12-13
Prevent sync chunk operations in sign system

### LINKED ISSUE #381: Sync chunk loading in cloudnet-modules/cloudnet-signs causes low TPS
**Description of the bug**
Component: **cloudnet-modules**/**cloudnet-signs**

When checking if a specific location is a sign and the chunk is not loaded, the loading is synchronous.

In my case, the **TPS dropped to 5 while CPU utilization rose to 80-90%** across all cores.

The root cause of the low performance was the synchronous call [Location#getBlock](https://github.com/CloudNetService/CloudNet-v3/blob/master/cloudnet-modules/cloudnet-signs/src/main/java/de/dytanic/cloudnet/ext/signs/bukkit/BukkitSignManagement.java#L55) in BukkitSignManagment.

I would be able to provide a matching PR if there is interest in fixing this.
The fix does require the 'Paper' API. Only loading chunks async when the required method is available could also be engineered (might be a bit more 'hacky').

My fix improves global performance by 1000% and plugin performance by about 950% (see **'Timings' Reports**).

**Steps to reproduce this bug**
I was not able to reproduce the bug on other Worlds apart from where it first occurred.

It seems like a lot of entities or strain in general needs to exist in unloaded chunks.

**My Fix**
Wrapping the [Location#getBlock](https://github.com/CloudNetService/CloudNet-v3/blob/master/cloudnet-modules/cloudnet-signs/src/main/java/de/dytanic/cloudnet/ext/signs/bukkit/BukkitSignManagement.java#L55) call with [World#getChunkAtAsync](https://papermc.io/javadocs/paper/1.15/org/bukkit/World.html#getChunkAtAsync-org.bukkit.Location-).

**Expected behavior**
In general the usage of Paper async chunk loading.
For this case, no game-breaking performance.

**'Timings' Reports**
[Befor My Fix](https://timings.aikar.co/?id=08f8ba7039134153b64c2d86768d80ff#plugins)
[After My Mix](https://timings.aikar.co/?id=d78a9ba2e9b04d3695ee603d19164e79#plugins)

**Environment and used hardware**
 System specifications: VPS, 6 cores, Intel® Xeon® E5-2620v3 / E5-2630v4 / 4114 / AMD EPYC™ 7282 (one of these)
 OS: Ubuntu 18.04 LTS
 Version: CloudNet 3.3.0

derklaro:
This pull request includes:

- [ ] breaking changes
- [X] no breaking changes

This pull request prevents the sync load and generate operations to chunks in which cloudsigns are created.

https://timings.aikar.co/?id=c23f0ed967384c95abc1478986be679d#plugins

Closes #381

### DIFF
--- a/cloudnet-modules/cloudnet-signs/src/main/java/de/dytanic/cloudnet/ext/signs/bukkit/BukkitSignManagement.java
+++ b/cloudnet-modules/cloudnet-signs/src/main/java/de/dytanic/cloudnet/ext/signs/bukkit/BukkitSignManagement.java
@@ -11,10 +11,12 @@ import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.Location;
 import org.bukkit.Material;
+import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
 import org.bukkit.material.MaterialData;
+import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -22,11 +24,11 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Iterator;
 
-public final class BukkitSignManagement extends AbstractSignManagement {
+public class BukkitSignManagement extends AbstractSignManagement {
 
     private final BukkitCloudNetSignsPlugin plugin;
 
-    BukkitSignManagement(BukkitCloudNetSignsPlugin plugin) {
+    protected BukkitSignManagement(BukkitCloudNetSignsPlugin plugin) {
         super();
 
         this.plugin = plugin;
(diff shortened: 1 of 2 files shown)


cloudnet-modules/cloudnet-signs/src/main/java/de/dytanic/cloudnet/ext/signs/bukkit/BukkitSignManagement.java
```diff
@@ -172,17 +164,11 @@ private BlockFace getSignFacing(BlockState blockState) {
     public Location toLocation(WorldPosition worldPosition) {
         Preconditions.checkNotNull(worldPosition);
 
-        return Bukkit.getWorld(worldPosition.getWorld()) != null ? new Location(
-                Bukkit.getWorld(worldPosition.getWorld()),
-                worldPosition.getX(),
-                worldPosition.getY(),
-                worldPosition.getZ()
-        ) : null;
+        final World world = Bukkit.getWorld(worldPosition.getWorld());
```
juliarn: Same as above.

juliarn (changes requested):
Looks good, but we should check if these changes can be applied to the NukkitSignManagement too.

cloudnet-modules/cloudnet-signs/src/main/java/de/dytanic/cloudnet/ext/signs/bukkit/BukkitSignManagement.java
```diff
@@ -40,28 +42,27 @@
...
-
-            if (location != null) {
-
-                Block block = location.getBlock();
-
-                if (block.getState() instanceof org.bukkit.block.Sign) {
-                    org.bukkit.block.Sign bukkitSign = (org.bukkit.block.Sign) block.getState();
-
-                    this.updateSign(location, sign, bukkitSign, signLayout, serviceInfoSnapshot);
+            final Location location = this.toLocation(sign.getWorldPosition());
```
juliarn: We don't use final for local variables at any place in this project, so why would we do this here? Imo final on local variables is pretty unnecessary.

einrobin (approved):
Didn't test if it works though

juliarn (approved):

PR #386, opened 2020-12-14
Fixed Endless kick Loop in Velocity Bridge Implementation

davidmayr:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

I fixed the bug that I documented here: https://discord.com/channels/325362837184577536/536594898321670154/788033187770400819

For Users that aren't on the Discord/don't speak German:

If you connect to Velocity on a 1.16 network in an older version of Minecraft (1.8 for example) the console is spammed in CloudNet, Velocity and the target server. The server always kicks the player for a too old version and your Velocity plugin sends the player back to the specific server that kicked the player.... This causes an endless loop until disconnection.

What I have fixed:
I compared the Bungeecord and the Velocity implementations and implemented the missing lines in the Velocity implementation.

I tested with one and two Lobbys. Everything works fine.

With 2 Lobbys:
CloudNet seems to send the player twice to Lobby-1 and once to Lobby-2 and then: KICK. Works perfectly. No endless loop.

With 1 Lobby:
It again sends the player twice to Lobby-1 and then kicks him.

https://discord.com/channels/325362837184577536/536594898321670154/788033187770400819

### DIFF
--- a/cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/VelocityCloudNetHelper.java
+++ b/cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/VelocityCloudNetHelper.java
@@ -90,10 +90,11 @@ public final class VelocityCloudNetHelper {
                 BridgeHelper.createOwnNetworkServiceInfo()
         );
     }
-    public static Optional<RegisteredServer> getNextFallback(Player player, RegisteredServer registeredServer) {
+
+    public static Optional<RegisteredServer> getNextFallback(Player player) {
         return BridgeProxyHelper.getNextFallback(
                 player.getUniqueId(),
-                registeredServer != null ? registeredServer.getServerInfo().getName() : null,
+                player.getCurrentServer().map(ServerConnection::getServerInfo).map(ServerInfo::getName).orElse(null),
                 player::hasPermission
         ).map(serviceInfoSnapshot -> new VelocityPlayerFallbackEvent(player, serviceInfoSnapshot, serviceInfoSnapshot.getName()))
                 .map(event -> proxyServer.getEventManager().fire(event))
@@ -107,11 +108,6 @@ public final class VelocityCloudNetHelper {
                 })
                 .map(VelocityPlayerFallbackEvent::getFallbackName)
                 .flatMap(proxyServer::getServer);
-    }
-
-    public static Optional<RegisteredServer> getNextFallback(Player player) {
-        return getNextFallback(player, player.getCurrentServer()
-                .map(ServerConnection::getServer).orElse(null));
     }
 
     public static CompletableFuture<ServiceInfoSnapshot> connectToFallback(Player player, String currentServer) {
--- a/cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/VelocityCloudNetHelper.java
+++ b/cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/VelocityCloudNetHelper.java
@@ -90,11 +90,10 @@ public final class VelocityCloudNetHelper {
                 BridgeHelper.createOwnNetworkServiceInfo()
         );
     }
-
-    public static Optional<RegisteredServer> getNextFallback(Player player) {
+    public static Optional<RegisteredServer> getNextFallback(Player player, RegisteredServer registeredServer) {
         return BridgeProxyHelper.getNextFallback(
                 player.getUniqueId(),
-                player.getCurrentServer().map(ServerConnection::getServerInfo).map(ServerInfo::getName).orElse(null),
+                registeredServer != null ? registeredServer.getServerInfo().getName() : null,
                 player::hasPermission
         ).map(serviceInfoSnapshot -> new VelocityPlayerFallbackEvent(player, serviceInfoSnapshot, serviceInfoSnapshot.getName()))
                 .map(event -> proxyServer.getEventManager().fire(event))
@@ -108,6 +107,11 @@ public final class VelocityCloudNetHelper {
                 })
                 .map(VelocityPlayerFallbackEvent::getFallbackName)
                 .flatMap(proxyServer::getServer);
+    }
+
+    public static Optional<RegisteredServer> getNextFallback(Player player) {
+        return getNextFallback(player, player.getCurrentServer()
+                .map(ServerConnection::getServer).orElse(null));
     }
 
     public static CompletableFuture<ServiceInfoSnapshot> connectToFallback(Player player, String currentServer) {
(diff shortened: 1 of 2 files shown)


cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
```diff
@@ -72,8 +72,14 @@ public void handle(ServerConnectedEvent event) {
 
     @Subscribe
     public void handle(KickedFromServerEvent event) {
-        VelocityCloudNetHelper.getNextFallback(event.getPlayer())
+        if(!event.getPlayer().isActive())
```
derklaro: Please don't return here, just do 
```java
if (event.getPlayer().isActive()) {
  // code
}
```

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
```diff
@@ -72,8 +72,14 @@ public void handle(ServerConnectedEvent event) {
...
+            return;
+
+        BridgeProxyHelper.handleConnectionFailed(event.getPlayer().getUniqueId(), event.getServer().getServerInfo().getName());
+
+        VelocityCloudNetHelper.getNextFallback(event.getPlayer(), event.getServer())
```
derklaro: As you called `handleConnectionFailed` there is no need of passing the server to the method, it simply generates an overload

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
```diff
@@ -72,8 +72,14 @@ public void handle(ServerConnectedEvent event) {
 
     @Subscribe
     public void handle(KickedFromServerEvent event) {
-        VelocityCloudNetHelper.getNextFallback(event.getPlayer())
+        if(!event.getPlayer().isActive())
+            return;
+
+        BridgeProxyHelper.handleConnectionFailed(event.getPlayer().getUniqueId(), event.getServer().getServerInfo().getName());
+
```
derklaro: No need for the empty line

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
```diff
@@ -72,8 +72,14 @@ public void handle(ServerConnectedEvent event) {
...
+        VelocityCloudNetHelper.getNextFallback(event.getPlayer(), event.getServer())
                 .ifPresent(registeredServer -> event.setResult(KickedFromServerEvent.RedirectPlayer.create(registeredServer)));
+
```
derklaro: Same here

derklaro (changes requested):
LGTM but please take care of the following things:

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/VelocityCloudNetHelper.java
```diff
@@ -90,11 +90,10 @@ public static NetworkConnectionInfo createNetworkConnectionInfo(Player player) {
                 BridgeHelper.createOwnNetworkServiceInfo()
         );
     }
-
-    public static Optional<RegisteredServer> getNextFallback(Player player) {
+    public static Optional<RegisteredServer> getNextFallback(Player player, RegisteredServer registeredServer) {
         return BridgeProxyHelper.getNextFallback(
                 player.getUniqueId(),
-                player.getCurrentServer().map(ServerConnection::getServerInfo).map(ServerInfo::getName).orElse(null),
```
juliarn: Everything in this class should be fine, why did you change this?

davidmayr: This actually fixed it for a single Lobby. This was my first try.

I guess I will revert this then. But why is it in the Bungeecord implementation when it shouldn't be here?

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
davidmayr: (Copy and Paste from my Comment to @juliarn's review) 
This actually fixed it for a single Lobby. This was my first try.

I guess I will revert this then. But why is it in the Bungeecord implementation when it shouldn't be here?

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/VelocityCloudNetHelper.java
juliarn: Because the velocity api returns optionals, which make a way cleaner api usage possible. Bungee is not doing this, so the code is a bit different.

davidmayr: Well I changed it because the current server was null but I think with handleConnectionFailed I can just ignore this. Reverting this now.

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
davidmayr: What empty line? The first or the second?

Since I did some changes now could you review them too? Is it fine now?

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
```diff
@@ -72,15 +72,13 @@ public void handle(ServerConnectedEvent event) {
 
     @Subscribe
     public void handle(KickedFromServerEvent event) {
-        if(!event.getPlayer().isActive())
-            return;
+        if(event.getPlayer().isActive()) {
```
derklaro: Can you please reformat to `if (..`?

cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/velocity/listener/VelocityPlayerListener.java
derklaro: Looks good now

derklaro (approved):
LGTM now

juliarn (approved):

juliarn:
Thank you

PR #417, opened 2021-04-05
Fix NPE when checking velocity console permissions

WebbiiiCC:
This pull request includes:

- [ ] breaking changes
- [X] no breaking changes

Return a permission function even when the PermissionSubject is not a player to prevent errors when executing some commands from the console in Velocity


derklaro:
The provider is only set for players in the setup event...

WebbiiiCC:
There is no check anywhere checking if the PermissionSubject really is a player. PermissionsSetupEvent is also called once for the velocity console so it has to be checked. Either we return a valid PermissionFunction or we add a check to the Event Listener

juliarn:
Its checked in the event, so should be fine
image

WebbiiiCC:
Doesn't show that to me...
https://github.com/CloudNetService/CloudNet-v3/blob/master/cloudnet-modules/cloudnet-cloudperms/src/main/java/de/dytanic/cloudnet/ext/cloudperms/velocity/listener/VelocityCloudNetCloudPermissionsPlayerListener.java#L34

juliarn:
This is on the master branch, we fixed that on development

WebbiiiCC:
Oh i see. Then sorry, my mistake.

juliarn:
No problem, thank your for the report anyway

PR #419, opened 2021-04-07
Set max-player property on proxies too

0utplay:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Added the Max-Players property also to proxies

### DIFF
--- a/cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/bungee/BungeeCloudNetBridgePlugin.java
+++ b/cloudnet-modules/cloudnet-bridge/src/main/java/de/dytanic/cloudnet/ext/bridge/bungee/BungeeCloudNetBridgePlugin.java
@@ -28,6 +28,7 @@ public final class BungeeCloudNetBridgePlugin extends Plugin {
     public synchronized void onEnable() {
         CloudNetDriver.getInstance().getServicesRegistry().registerService(IPlayerManager.class, "BridgePlayerManager", new BridgePlayerManager());
 
+        BungeeCloudNetHelper.init();
         this.initListeners();
         this.registerCommands();
         this.initServers();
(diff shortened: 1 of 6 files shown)


juliarn:
This might require a bit more handling, as the max players could be changed in the proxy ping event. We handle that for servers, so it might be good for proxies too. Altough calling the ping event to get the max players is a bit weird.

Sarsum:
Well, we were thinking about the Ping Event. But that would cause problems with the SyncProxy Max Players..
The initial idea was to add this, to support Proxies with the SmartModule (which is not working right now, an the max player count is always 0 in the console)

derklaro (approved):
LGTM

juliarn:
> Well, we were thinking about the Ping Event. But that would cause problems with the SyncProxy Max Players..
> The initial idea was to add this, to support Proxies with the SmartModule (which is not working right now, an the max player count is always 0 in the console)

That would not cause problems with SyncProxy, it's the other way around: CloudNet would detect and use the max player amount set by SyncProxy, which is the correct behavior. The problem with this currently is, that the max players in the bungee config are mostly ignored and replaced via the ping event. If CloudNet would call that event, the same way it does on servers, everything should be fine.

juliarn:
What CloudNet could do too is to add an event handler with highest priority to get the max players from the event. That would work if no other plugin modifies the max players in this priority.

Sarsum:
> 
> 
> > Well, we were thinking about the Ping Event. But that would cause problems with the SyncProxy Max Players..
> > The initial idea was to add this, to support Proxies with the SmartModule (which is not working right now, an the max player count is always 0 in the console)
> 
> That would not cause problems with SyncProxy, it's the other way around: CloudNet would detect and use the max player amount set by SyncProxy, which is the correct behavior. The problem with this currently is, that the max players in the bungee config are mostly ignored and replaced via the ping event. If CloudNet would call that event, the same way it does on servers, everything should be fine.

This would not work in the correct way with the SmartModule. When the amount of the SyncProxy Module is used (e.g. 1000) the SmartModule would use this global amount for starting a new proxy. (And not for example 100 for each Proxy)

Sarsum:
~~I just got the following idea:~~
~~We could handle the "Max-players" like you @juliarn described and add another number to the Smart Configuration.~~
~~I would call it something like `customUsedMaxPlayerAmountForCalculations` (sorry, I'm bad with picking names for variables). If its `-1` (default), just ignore it and use "Max-players", but if its not, use this number instead of the "Max-players".~~

~~I first thought of adding this new variable to the SyncProxy Config, but then the SmartModule would somehow depend on the SyncProxy Module.~~

Sarsum:
Nevermind my previous idea, this would prevent the dynamic change of the max players with the API (or the max players used by the smart module)...

juliarn (approved):

PR #433, opened 2021-04-11
Fix service state when stopping a service

derklaro:
This pull request includes:

- [ ] breaking changes
- [X] no breaking changes

When stopping a service using `ser NAME stop` and `autoDeleteOnStop` is set to `false` the service will rest in the STOPPED state rather than changing back the prepared state. This is the only logic thing to do, as you can now see which services were stopped and which services were prepared. This may help the user so they don't need to take notes which services got stopped by them and which got created.

Looks fine for now.

%


juliarn:
In my opinion, this is completely unnecessary and confusing for the user. There is no real benefit and there is one state more to handle. Changing this behavior might break the code of api users. And why would you separate this in two states if it's basically the same and should be handled the same?

derklaro:
It's the whole point: prepared != stopped... When i stop a server there are old files which the user may need, the worlds etc. The stopped state is just an indication that the service got stopped, prepared is a clean and fresh service

PR #437, opened 2021-04-12
Add check if you really can bind on the given ips in setup

### LINKED ISSUE #352: Test whether the IP/port provided in the setup can really be bound on

0utplay:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Add a check in the setup whether really can be bound on the ip and port

Tested the initial setup & it works a intended

Fixes #352

### DIFF
--- a/cloudnet/src/main/java/de/dytanic/cloudnet/console/animation/questionlist/answer/QuestionAnswerTypeHostAndPort.java
+++ b/cloudnet/src/main/java/de/dytanic/cloudnet/console/animation/questionlist/answer/QuestionAnswerTypeHostAndPort.java
@@ -35,7 +35,7 @@ public class QuestionAnswerTypeHostAndPort implements QuestionAnswerType<HostAnd
             URI uri = URI.create("tcp://" + input);
 
             String host = uri.getHost();
-            if (host == null || (requiresPort && uri.getPort() == -1)) {
+            if (host == null || (this.requiresPort && uri.getPort() == -1)) {
                 return null;
             }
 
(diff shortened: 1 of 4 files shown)


cloudnet/src/main/java/de/dytanic/cloudnet/console/animation/questionlist/answer/QuestionAnswerTypeHostAndPort.java
```diff
@@ -24,11 +35,14 @@ public boolean isValidInput(@NotNull String input) {
             URI uri = URI.create("tcp://" + input);
 
             String host = uri.getHost();
-            if (host == null) {
+            if (host == null || (requiresPort && uri.getPort() == -1)) {
```
derklaro: this.requiresPort

derklaro (changes requested):

derklaro (approved):

cloudnet/src/main/java/de/dytanic/cloudnet/console/animation/questionlist/answer/QuestionAnswerTypeHostAndPort.java
```diff
@@ -24,11 +35,14 @@ public boolean isValidInput(@NotNull String input) {
             URI uri = URI.create("tcp://" + input);
 
             String host = uri.getHost();
-            if (host == null) {
+            if (host == null || (this.requiresPort && uri.getPort() == -1)) {
                 return null;
             }
 
             InetAddress inetAddress = InetAddresses.forUriString(host);
+            if (this.requiresPort && !PortValidator.checkHost(inetAddress.getHostAddress(), uri.getPort())) {
```
juliarn: I think we should check the host too if the port is not required. This might need a new method in the PortValidator - when address already in use or binding success return true otherwise return false.

juliarn (approved):

PR #448, opened 2021-04-24
Switch to a fail-over cluster solution

derklaro:
This pull request includes:

- [ ] breaking changes
- [X] no breaking changes

CloudNet's current cluster solution works based on a cluster solution that is not widely used. Each node in the cluster decides itself which services should be started and how the services should be named. This is one of the biggest problems of CloudNet at the moment. Different nodes start a service of a certain task at the same time and create two services with exactly the same name.

unknown

The head node of the fail-over cluster is supposed to solve exactly this problem. A node in the network decides which nodes should start which services and how they should be named. To prevent nodes in the cluster that are not connected but should start services, a node must send an update of its info at least every 30 seconds to stay connected. If the head node fails, the "oldest" node (the node that was started first) takes over the job of the head node. Duplicate services do not occur after this change. Furthermore, this change contains an attempted fix to prevent "ghost" services, i.e. services that are magically still registered as STOPPED on other nodes even though they have long since been unregistered.

The changes were tested by us with 3 nodes and by a user with 3+ nodes and seem to work (at least there were no problems with 300 and more players)


juliarn (approved):
finally :peepoHappy:

PR #456, opened 2021-04-27
Use gradle wrapper instead of gradle provided by the ci

derklaro:
This pull request includes:

- [ ] breaking changes
- [X] no breaking changes

The Jenkins is now using the gradle wrapper from the repository. Some stages were merged together.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,8 +12,8 @@ pipeline {
 
     stages {
         stage('Clean-Build') {
-            configFileProvider([configFile(fileId: "e94f788c-1d9c-48d4-b9a9-8286ff68275e", targetLocation: 'gradle.properties')]) {
-                steps {
+            steps {
+                configFileProvider([configFile(fileId: "e94f788c-1d9c-48d4-b9a9-8286ff68275e", targetLocation: 'gradle.properties')]) {
                     sh './gradlew clean build jar -x test -x allJavaDoc --full-stacktrace'
                 }
             }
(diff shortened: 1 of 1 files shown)


juliarn (changes requested):
The daemon is already disabled by the gradle.properties file which is inserted by the Jenkins, so this is not necessary via the commands. Doing it like this will make things easier if we would want to enable it again for example. You actually removed the insertion of the gradle.properties. I would also put the test into a separate step after the clean build.

derklaro:
> The daemon is already disabled by the gradle.properties file which is inserted by the Jenkins, so this is not necessary via the commands. Doing it like this will make things easier if we would want to enable it again for example. You actually removed the insertion of the gradle.properties. I would also put the test into a separate step after the clean build.

* I think it's better to disable the gradle deamon manually or to put the configuration file into the repository directly because it's much easier to understand/reproduce how the build lifecycle work on the ci

* I think there is no reason the run the tests in an extra step as they are running during the build anyways. We are just wasting time when we are doing that...

juliarn:
Why are the tests running anyway? We can run them seperately after the build. Putting them in a seperate step shows better where the CI failed for example, it also allows us to skip tests etc.
I understand that its harder to get what is happening if we manage gradle via the gradle.properties file, but this file is specified for our CI and it's way easier for us to change something in the Jenkins configuration than via GitHub. There are also other things like the repo credentials in the gradle.properties, so we have to append it anyway.

juliarn (approved):

PR #496, opened 2021-06-20
SSH Key support for the SFTP TemplateStorage

einrobin:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Added SSH key support to the SFTP TemplateStorage module, the password authentication is still available.


Sarsum:
How would it be to add another boolean to the config to support password encrypted SSH-Keys? If it is turned on, you can pass the password into the JSch addIdentify method

derklaro (commented):
Looks fine!

PR #523, opened 2021-07-26
Create migrate command for migration between supported databases

0utplay:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Created a command to migrate the different databases.

Tested the command with all supported databases with 100.000 entries

### DIFF
--- a/cloudnet/src/main/resources/files/versions.json
+++ b/cloudnet/src/main/resources/files/versions.json
@@ -163,7 +163,7 @@
         {
           "name": "latest",
           "cacheFiles": false,
-          "url": "https://ci.codemc.io/job/Spottedleaf/job/Tuinity/lastSuccessfulBuild/artifact/tuinity-paperclip.jar",
+          "url": "https://ci.codemc.io/job/Spottedleaf/job/Tuinity-1.17/lastSuccessfulBuild/artifact/tuinity-paperclip.jar",
           "minJavaVersion": 16,
           "properties": {
             "copy": {
@@ -176,7 +176,7 @@
         },
         {
           "name": "1.17.1",
-          "url": "https://ci.codemc.io/job/Spottedleaf/job/Tuinity/lastSuccessfulBuild/artifact/tuinity-paperclip.jar",
+          "url": "https://ci.codemc.io/job/Spottedleaf/job/Tuinity-1.17/lastSuccessfulBuild/artifact/tuinity-paperclip.jar",
           "minJavaVersion": 16,
           "properties": {
             "copy": {


cloudnet/src/main/java/de/dytanic/cloudnet/command/commands/CommandMigrate.java
```diff
@@ -0,0 +1,119 @@
...
+
+            if (sourceDatabaseProvider.equals(targetDatabaseProvider)) {
+              sender.sendMessage(LanguageManager.getMessage("command-migrate-source-equals-target"));
+              return;
+            }
+
+            executeIfNotCurrentProvider(sourceDatabaseProvider, AbstractDatabaseProvider::init);
+            executeIfNotCurrentProvider(targetDatabaseProvider, AbstractDatabaseProvider::init);
+
+            Integer chunkSize = Ints.tryParse(properties.getOrDefault("chunk-size", Integer.toString(100)));
```
juliarn: Create a constant DEFAULT_CHUNK_SIZE

cloudnet/src/main/java/de/dytanic/cloudnet/command/commands/CommandMigrate.java
```diff
@@ -0,0 +1,119 @@
...
+    return CloudNet.getInstance().getServicesRegistry().getService(AbstractDatabaseProvider.class, name);
+  }
+
+  private static void executeIfNotCurrentProvider(@NotNull AbstractDatabaseProvider provider,
+    @NotNull ThrowableConsumer<AbstractDatabaseProvider, ?> handler) {
+    if (!CloudNet.getInstance().getDatabaseProvider().equals(provider)) {
+      try {
+        handler.accept(provider);
+      } catch (Throwable throwable) {
+        throwable.printStackTrace();
```
juliarn: Don’t catch Throwables, only catch Exceptions as Errors would get catched too. Also this should be properly logged or rethrown.

cloudnet/src/main/java/de/dytanic/cloudnet/command/commands/CommandMigrate.java
```diff
@@ -0,0 +1,119 @@
...
+    super.prefix = "cloudnet";
+    super.permission = "cloudnet.command." + super.names[0];
+    super.description = LanguageManager.getMessage("command-description-migrate");
+  }
+
+  private static Predicate<String> providerRegisteredPredicate() {
+    return input -> getDatabaseProvider(input) != null;
+  }
+
+  private static Supplier<Collection<String>> possibleAnswerSupplier() {
```
juliarn: possibleDatabaseProvidersSupplier() would be a more matching name

cloudnet/src/main/java/de/dytanic/cloudnet/command/commands/CommandMigrate.java
```diff
@@ -0,0 +1,119 @@
...
+              sender.sendMessage(LanguageManager.getMessage("command-migrate-source-equals-target"));
+              return;
+            }
+
+            executeIfNotCurrentProvider(sourceDatabaseProvider, AbstractDatabaseProvider::init);
+            executeIfNotCurrentProvider(targetDatabaseProvider, AbstractDatabaseProvider::init);
+
+            Integer chunkSize = Ints.tryParse(properties.getOrDefault("chunk-size", Integer.toString(100)));
+            if (chunkSize == null) {
+              chunkSize = 100;
```
juliarn: Telling the user that the input wasn’t a number might be better than just using the default.

juliarn (changes requested):
Looks good, just a few suggestions for improvement. Also failure of migration is not really handled, for example when the connection to the database fails etc. Would be good if we would add that and print a understandable message for the user if this is the case.

PR #541, opened 2021-08-12
Priority for Templates

0utplay:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Added a priority option for templates. Using the priority, it's possible to specify which template gets copied first and which last.

Tested starting the service, set different priorities for group and task templates.

Adds support for #333

### DIFF
--- a/cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/service/ServiceTemplate.java
+++ b/cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/service/ServiceTemplate.java
@@ -144,7 +144,8 @@ public class ServiceTemplate implements INameable, SerializableObject, Comparabl
 
   /**
    * This priority is used to determine in which order the template should be installed e.g. a Template with a priority
-   * of 1 (high prio) is installed after Templates with a lower prio (e.g. 5) are installed
+   * of 10 (high prio) is installed after Templates with a lower prio (e.g. 1) are installed to prevent that Templates
+   * with lower priorities overwrite Templates with a higher one
    *
    * @return the priority of the template
    */
@@ -180,7 +181,7 @@ public class ServiceTemplate implements INameable, SerializableObject, Comparabl
     buffer.writeString(this.name);
     buffer.writeString(this.storage);
     buffer.writeBoolean(this.alwaysCopyToStaticServices);
-    buffer.writeInt(this.priority);
+    buffer.writeVarInt(this.priority);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/service/ServiceTemplate.java
```diff
@@ -164,6 +180,7 @@ public void write(@NotNull ProtocolBuffer buffer) {
     buffer.writeString(this.name);
     buffer.writeString(this.storage);
     buffer.writeBoolean(this.alwaysCopyToStaticServices);
+    buffer.writeInt(this.priority);
```
derklaro: I think we can use a var int here to save bytes

cloudnet/src/main/java/de/dytanic/cloudnet/service/defaults/DefaultTemplateCloudService.java
```diff
@@ -151,8 +151,9 @@ private boolean includeInclusions0(ServiceRemoteInclusion inclusion, Path destin
 
   @Override
   public void includeTemplates() {
+    Collections.sort(this.waitingTemplates);
     while (!this.waitingTemplates.isEmpty()) {
-      ServiceTemplate template = this.waitingTemplates.poll();
+      ServiceTemplate template = this.waitingTemplates.remove(0);
```
derklaro: Why is this still a while? We can use a for loop here

cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/service/ServiceTemplate.java
```diff
@@ -136,6 +142,16 @@ public String getStorage() {
     return this.storage;
   }
 
+  /**
+   * This priority is used to determine in which order the template should be installed e.g. a Template with a priority
+   * of 1 (high prio) is installed after Templates with a lower prio (e.g. 5) are installed
```
derklaro: Is this documentation comment correct? Because why is 1 a "high priority" when its installed AFTER a low priority 5?

derklaro (changes requested):

cloudnet-driver/src/main/java/de/dytanic/cloudnet/driver/service/ServiceTemplate.java
0utplay: Yes, the Template with a higher prio is not overwritten by the template with the lower prio. That's why 1 is a high prio and gets installed after

derklaro: Then explain it in this way but in the current way its just confusing

0utplay: Actually, I'll change the sorting so that 0 has a low prio and e.g. 100 has a high one to prevent that new templates break the prio

cloudnet/src/main/java/de/dytanic/cloudnet/service/defaults/DefaultTemplateCloudService.java
0utplay: I don't see a problem using a while here. Using a for would result in a extra variable because the size of the list changes after each element

derklaro (approved):

PR #572, opened 2021-12-27
Dockerized services

derklaro:
This pull request introduces support for services to run in docker containers. The user is free to configure each setting the underlying docker-java libarary can support. Configuration can be done via the module configuration or (for some container-specific options) via the service task properties (These options are currently: additional volumes, additional exposed ports & specific java docker image (defaults to `azul/zulu-openjdk-alpine:17-jre`)).


0utplay (approved):
LGTM

PR #573, opened 2021-12-28
Handle command exceptions for websocket messages as well

PierreSchwang:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Command-Exceptions thrown by cloud are handled in the native console by catching every exception type and printing a human-readable message to the user. When sending commands using the WebSocket connection, the thrown exceptions are wrapped by a CompletionException and the full stack trace is printed to sout (see Before-image).
Now the CompletionException and their underlying Exception is handled and **not** thrown if the Exception is handled by CommandExceptionHandler.

Before: grafik
After: grafik

N/A

### DIFF
--- a/modules/rest/src/main/java/de/dytanic/cloudnet/ext/rest/v2/V2HttpHandlerNode.java
+++ b/modules/rest/src/main/java/de/dytanic/cloudnet/ext/rest/v2/V2HttpHandlerNode.java
@@ -16,6 +16,7 @@
 
 package de.dytanic.cloudnet.ext.rest.v2;
 
+import cloud.commandframework.exceptions.CommandParseException;
 import de.dytanic.cloudnet.CloudNet;
 import de.dytanic.cloudnet.common.log.AbstractHandler;
 import de.dytanic.cloudnet.common.log.LogManager;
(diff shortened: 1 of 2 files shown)


0utplay:
Thanks for the PR!
I think we need to take a closer look into this as these exceptions occur on other API calls too. 
So we need to change it in the backing api rather than changing it at each usage.

PierreSchwang:
Changed the core functionality of the DefaultCommandProvider. As exceptions are logged in the CommandExceptionHandler, all exceptions thrown somewehere in the CompletableFuture-Chain are now suppressed. That way all usages should be covered. 

I kept the format of the websocket messages, as  otherwise that would differ from the default command log.

derklaro:
I don't think that removing the exception from the returned future is a good way to handle this, as one of the main reason why this is a CompletableFuture is that api users can actually get the exception if needed. There are some other points in the code which are using `join` as well, but I think at the web socket part we can just use `thenRun` and send all messages back to the user? I don't see the reason currently why we should block there anyways.

0utplay:
We've addressed the issue internally, thanks for your contribution anyways.

PR #577, opened 2022-01-04
docs: document all channel message classes

derklaro:


driver/src/main/java/eu/cloudnetservice/cloudnet/driver/channel/ChannelMessage.java
```diff
@@ -37,40 +75,111 @@ public record ChannelMessage(
...
+   */
   @Contract("_ -> new")
   public static @NonNull Builder buildResponseFor(@NonNull ChannelMessage input) {
     return builder().channel("").message("").target(input.sender.type(), input.sender.name());
   }
 
+  /**
+   * Sends this channel message using the current messenger of the environment. This is in fact just a shortcut method
+   * for {@link CloudMessenger#sendChannelMessage(ChannelMessage)}. This method will not wait for the target component
+   * to respond (it doesn't event expects a response) but for the handling component to send the message.
```
0utplay: Something is wrong here.

-> `it doesn't expect a response`

0utplay (changes requested):
See above

PR #591, opened 2022-01-17
Remove unsafe socket address casts

derklaro:
Especaially for bungeecord, as noted in the documentation you can connect to the proxy using a domain socket address which is not an InetSocketAddress.

We cannot move the netty casts to the HostAndPort class directly because of the relocation happening during the compile.


0utplay (approved):
LGTM

PR #605, opened 2022-01-28
Add support to auto generate rpc based api classes

derklaro:
This pr adds support to auto generate rpc classes. The classes will be auto generated and contain (by default, can be changed) only methods which are abstract and downstream the invocations of them directly into a rpc sender. Class extending and retreiving of the used rpc sender is possible as well. 

See the added tests for further information how this works :)


0utplay (approved):
LGTM

PR #608, opened 2022-02-04
Improve the handling of nodes in the cluster

derklaro:
This pull request improves the handling of nodes within a cluster, api wise and internally. 

This includes a better disconnect handling, allowing nodes to reconnect within a given time when marked as "disconnected". This behaviour is disabled by default as (also explained in the launcher.cnl file) this might in some cases lead to unexpected data sync behaviour in the cluster making it unusable. (For example players are still marked as connected which are gone already, services are registered while unregistered on other nodes etc.) The local node, if elected as the head node will stop starting services if any node in the cluster is marked locally as disconnected. This is required to ensure that not multiple nodes are taking the head status and begin to start services which are namewise duplicates. 

**The disconnect handling might not be working correctly at all when scaling largely, use that setting at your own risk!** (might be a good note for the docs as well)


0utplay (approved):
lgtm

PR #609, opened 2022-02-04
Add influx db support

derklaro:
This pull request adds support for influxdb. This can be used to (for example) create a CloudNet dashboard with grafana.


0utplay (approved):
lgtm

PR #611, opened 2022-02-05
Add Available Servers and Max Servers in CloudNet NPC Module

EinfachBeez:
This pull request includes:

- [ ] breaking changes
- [x] no breaking changes

Gives you the possibility to use Available Servers and Max Servers in the Info Line. Available Server = a_s and Max Servers = m_s. I think the feature is actually quite practical, but I didn't find enough time to calculate the Max Services and therefore I take the data directly from the SmartConfig. But feel free to change it if there is a better way.

There are no errors or warnings. I've tested it directly on my server. 
TestonServer

I didn't think so :thinking:


derklaro:
Thanks for your pull request! I just want to point out that we're currently **not** accepting pull requests targeting development. If you want to add features to CloudNet, consider targeting your pull requests against `unstable/rewrite`, make sure you're following the style guidelines of the project (check if checkstyle passes during compile). Also your approach of getting the task by the target group name is just wrong. There is a **BIG** difference between a task and a group! (And there is no guarantee that the task actually has a smart configuration).

EinfachBeez:
That is also the reason why I meant that you can gladly change that. For my purpose it definitely worked, but if there is a need for this feature, you could always change it.

derklaro:
Closing for now because of the the already mentioned reasons.

PR #622, opened 2022-02-24
Document the whole driver service package

derklaro:


driver/src/main/java/eu/cloudnetservice/cloudnet/driver/service/ProcessSnapshot.java
```diff
@@ -21,13 +21,27 @@
...
+import java.util.Set;
 import lombok.NonNull;
 
 /**
- * A snapshot of a process in the Cloud which provides information about the cpu and memory usage, the running threads
- * and the pid.
+ * A snapshot of the process resources at a specific time. It holds the most useful information for displaying or
+ * storing statistics about a service / node (or anything else).
+ *
+ * @param pid                     the process id of component which created the snapshot.
```
0utplay:
```suggestion
 * @param pid                     the process id of the component which created the snapshot.
```

0utplay (changes requested):

PR #630, opened 2022-03-12
Update to netty 5

derklaro:
Netty 5 is expected to be released within the next few days. This pull request tracks the last changes they do before the release as well as making us ready to update asap.

Currently the following things are still missing in netty5 and we are waiting that they complete these:
 - [ ] SSL is not yet implemented for the new buffer api
 - [ ] ChunkedWriteHandler is not yet implemented for the new buffer api
 - [ ] io_uring is not yet updated to support netty 5
 
This pull request is currently untested in the real world and may be broken once using it to power CloudNet. Especially their new buffer api is a bit hard to keep track of as we are now required to do each step ourselfes (for example expanding the buffer before writing to it).

Note: Netty 5 is now enabled as the channel option `RCVBUF_ALLOCATOR_USE_BUFFER` is set to true. Reverting the change will break the network system completly as netty falls back to using ByteBufs (the old buffer api).


wrapper-jvm/build.gradle.kts
```diff
@@ -50,7 +45,7 @@ tasks.withType<ShadowJar> {
   doFirst {
     // Note: included dependencies will not be resolved, they must be available from the node resolution already
     from(exportLanguageFileInformation())
-    from(exportCnlFile("wrapper.cnl", ignoredGroupIds))
+    from(exportCnlFile("wrapper.cnl", ignoredGroupIds, false))
```
0utplay, line 48:
```suggestion
    from(exportCnlFile("wrapper.cnl", ignoredGroupIds, resolveRepos = false))
```

0utplay (changes requested):

0utplay (approved):

PR #646, opened 2022-03-27
fix: prevent runtime reloading of critical information

derklaro:
This for example prevents changing the node unique id by reloading the current node config which breaks the whole cluster completely. All API methods are still present and useable to do that anyway, but it's then up to the developers to ensure that nothing goes wrong when doing that.

### DIFF
--- a/node/src/main/java/eu/cloudnetservice/cloudnet/node/config/JsonConfiguration.java
+++ b/node/src/main/java/eu/cloudnetservice/cloudnet/node/config/JsonConfiguration.java
@@ -291,6 +291,39 @@ public final class JsonConfiguration implements Configuration {
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


node/src/main/java/eu/cloudnetservice/cloudnet/node/config/JsonConfiguration.java
```diff
@@ -290,6 +290,39 @@ public boolean fileExists() {
     return this;
   }
 
+  @Override
+  public void reloadFrom(@NonNull Configuration configuration) {
+    // supported identity changes
+    this.identity.properties().append(configuration.properties());
```
0utplay: Shouldn't we append properties of the identity?

0utplay (changes requested):

0utplay (approved):

PR #658, opened 2022-05-01
chore: cleanup and refactor sign code

derklaro:
Cleans up the sign system code by a lot to bring more consistency and easier readability. This should also improve the performance of the sign system, as for example copying the sign location to a platform location is no longer needed when updating the current shown lines of a sign.


modules/signs/src/main/java/eu/cloudnetservice/modules/signs/Sign.java
```diff
@@ -34,7 +30,7 @@
  */
 @ToString
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
-public class Sign implements Comparable<Sign> {
+public class Sign {
 
   public static final Type COLLECTION_TYPE = new TypeToken<Collection<Sign>>() {
```
TheMeinerLP, line 35: Can you change this to better way ?

modules/signs/src/main/java/eu/cloudnetservice/modules/signs/platform/sponge/SpongePlatformSign.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    // get the block at the given location
+    var entity = location.blockEntity().orElse(null);
+    if (entity instanceof org.spongepowered.api.block.entity.Sign sign) {
+      // set the glowing status if needed
+      sign.glowingText().set(layout.glowingColor() != null);
+
+      // set the sign lines
+      for (int i = 0; i < Math.min(4, layout.lines().size()); i++) {
```
TheMeinerLP, line 85: What happend if i set less as 4 lines in the config ?

derklaro: Well, Math.min returns the smaller of a and b - if you define more than 4 lines only 4 will be used, if you define less then that amount will be used instead.

TheMeinerLP: Okay perfect 👍🏻

modules/signs/src/main/java/eu/cloudnetservice/modules/signs/Sign.java
derklaro: Not without breaking compatibility with older spigot versions

TheMeinerLP: Can be make a issue.

0utplay (approved):
lgtm

PR #668, opened 2022-05-24
Add inclusions to deployments

0utplay:


node/src/main/java/eu/cloudnetservice/node/command/sub/ServiceCommand.java
```diff
@@ -419,9 +421,9 @@ private void displayServiceInfo(
 
   private @NonNull Collection<String> parseExcludes(@Nullable String excludes) {
     if (excludes == null) {
-      return Collections.emptyList();
+      return List.of();
     }
 
-    return Arrays.asList(excludes.split(";"));
+    return List.of(excludes.split(";"));
```
derklaro: I think we should replace that parsing logic with a guava splitter which can omit some common mistakes already:
`private static final Splitter SEMICOLON_SPLITTER = Splitter.on(';').omitEmptyStrings().trimResults();`

and this method like:
`return excludes == null ? List.of() : SEMICOLON_SPLITTER.splitToList(excludes);`

And as this method is now used multiple times for not only parsing excludes we should rename it and the parameter.

node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
```diff
@@ -498,9 +498,17 @@ protected void executeDeployment(@NonNull ServiceDeployment deployment) {
         var fileName = Files.isDirectory(path)
           ? path.getFileName().toString() + '/'
           : path.getFileName().toString();
-        // check if the file is ignored
-        return deployment.excludes().stream().noneMatch(pattern -> fileName.matches(pattern.replace("*", "(.*)")))
-          && !DEFAULT_DEPLOYMENT_EXCLUSIONS.contains(fileName);
+        // file matcher escaping asterix chars for regex
+        Predicate<String> fileMatcher = input -> fileName.matches(input.replace("*", "(.*)"));
```
derklaro: I think we can move that predicate into a static final variable. Maybe we should move the `*` -> `(.*)` replacement out of this method into the command and apply the pattern fixer to it there? (Can be done in a later step after this pr as well)

node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
```diff
@@ -498,9 +498,17 @@ protected void executeDeployment(@NonNull ServiceDeployment deployment) {
...
+        // if we have inclusions we just handle them, exclusions otherwise
+        if (deployment.includes().isEmpty()) {
+          // check if the file is excluded
+          return deployment.excludes().stream().noneMatch(fileMatcher)
+            && !DEFAULT_DEPLOYMENT_EXCLUSIONS.contains(fileName);
```
derklaro: Maybe an idea for later as well, but we could move the default exclusions into the service deployment builder and add a method to exclude them 🤔 rather than to enforce that exclusion...

driver/src/main/java/eu/cloudnetservice/driver/service/ServiceDeployment.java
```diff
@@ -171,6 +186,38 @@ public static class Builder {
...
+     */
+    public @NonNull Builder includes(@NonNull Collection<String> includes) {
+      this.includes = new HashSet<>(includes);
+      return this;
+    }
+
+    /**
+     * Adds the file names (in a regular expression form) which should be included when deploying a service. Directories
+     * must be suffixed with a {@literal /}.
+     * <p>
```
derklaro, line 235: We should include a paragraph here and on the `includes` method that explains that when exclusions are added to the deployment that excludes have no effect anymore, or preferability we should honor includes and excludes when provided 🤔

derklaro (changes requested):
Generally good, added some ideas (some of them for the future)

derklaro (approved):

PR #680, opened 2022-06-11
[ci skip] chore: require console sender for service screen command

derklaro:
### Motivation
The screen commands should not be available from the driver (or ingame) as the messages will not be send to the user who executed the command.

### Modification
The `service <name|uuid> toggle` annotation now has the required sender set to the console sender.

### Result
The `service <name|uuid> toggle` command is now only available from the node console.


schucko1337 (approved):
Nice work.

0utplay (approved):

PR #691, opened 2022-06-25
fix: use new cloud failure reason getter

derklaro:
### Motivation
In our fork of cloud we've exposed the failure reason when a flag parse exception is raised. That changed was later pr-ed to the upstream repository, but they use flow style for all new methods, so after updating our fork we now need to change this as well.

### Modification
Use new `FlagParseException#failureReason()` instead of `FlagParseException#getFailureReason()`.

### Result
The cloud update is completed and we can use the new, updated version of cloud.


0utplay (approved):
lgtm

schucko1337 (approved):

PR #692, opened 2022-06-27
chore: improve report module

derklaro:
### Motivation
The current report module is a bit messy and doesn't provide *all* the information which might be useful when it comes to support, but stores a lot of data which is in almost all cases not needed at all or doesn't provide any useful data for us to work with.

### Modification
The report module was rewritten with 3 big changes:
 - It now includes all the data which might be useful when it comes to supporting a user
 - It no longer stores the logs and information about services, which is normally not needed and only clutters your hard drive
 - Paste Servers can now be configured much easier, allowing to easily integrate custom paste servers

### Result
The report module is now much better when it comes down to useful support data and does no longer store cluttering data.


modules/report/src/main/java/eu/cloudnetservice/modules/report/command/ReportCommand.java
```diff
@@ -55,144 +57,189 @@ public final class ReportCommand {
...
+  @Parser(suggestions = "pasteServer")
+  public @NonNull PasteServer defaultPasteServerParser(
     @NonNull CommandContext<CommandSource> $,
     @NonNull Queue<String> input
   ) {
     var name = input.remove();
-    return this.reportModule.reportConfiguration().pasteServers()
+    return this.reportModule.configuration().pasteServers()
       .stream()
       .filter(service -> service.name().equalsIgnoreCase(name))
```
0utplay: service -> server

modules/report/src/main/java/eu/cloudnetservice/modules/report/emitter/SpecificReportDataEmitter.java
```diff
@@ -0,0 +1,52 @@
...
+
+public abstract class SpecificReportDataEmitter<T> implements ReportDataEmitter {
+
+  private final BiConsumer<ReportDataWriter, Collection<T>> sectionTitleDecorator;
+
+  public SpecificReportDataEmitter(@NonNull String sectionTitle) {
+    this((writer, ignored) -> writer.appendString(sectionTitle));
+  }
+
+  public SpecificReportDataEmitter(@NonNull BiConsumer<ReportDataWriter, Collection<T>> sectionTitleDecorator) {
```
0utplay, line 31: Nullability annotation

modules/report/src/main/java/eu/cloudnetservice/modules/report/emitter/defaults/NodeServerDataEmitter.java
```diff
@@ -0,0 +1,120 @@
...
+  @Override
+  public @NonNull ReportDataWriter emitData(@NonNull ReportDataWriter writer, @NonNull NodeServer value) {
+    // information about the node which is always present
+    writer = writer
+      .beginSection(titleWriter -> titleWriter
+        .appendString(value.name())
+        .appendString(" ")
+        .appendString(value.state().name())
+        .appendString(value.head() ? " (head)" : "")
+        .appendString(value instanceof LocalNodeServer ? " current" : " remote"))
```
0utplay, line 51: Maybe use local instead of current

0utplay (changes requested):

schucko1337 (approved):

0utplay (approved):

PR #696, opened 2022-07-02
feat: add support for annotated http handlers

derklaro:
### Motivation
Developing http handlers is currently painful, as handling a http request requires a seperate class per handler or one handler which does a lot of `if`s to check if a specific request path matches a handling path, if the http method matches etc. 

### Modification
Added the possibility to define multiple http handlers in the same class by annotating them and adding some useful functionality to them to make development even easier.

### Result
Develeoping http handlers is now much easiert.

##### Other context
All existing http handlers will be converted in a seperate pull request to not unnecessarily blow up this one.

Example usage of the new system:
```java
    @HttpRequestHandler(paths = "/test/all/{id}", priority = 0)
    public void handleAll(
      // the http call context (this is the only required paramater)
      HttpContext context,
      // this parameter is taken out of the path and required
      @RequestPathParam("id") String id,
      // this parameter is a required header
      @RequestHeader("username") String username,
      // this paramater is an optional header which defaults to null
      @Optional @RequestHeader("pageId") String pageId,
      // this paramater is an optional header, which defaults to "bcc582ed494d4b9386cbb58564651a26" if not given
      @Optional @RequestHeader(value = "uuid", def = "bcc582ed494d4b9386cbb58564651a26") String uuid
    ) {
    }
```


schucko1337 (approved):

driver/src/main/java/eu/cloudnetservice/driver/network/http/annotation/parser/HttpAnnotationProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+ * A processor for an annotation which can be added to a http handler parameter.
+ *
+ * @since 4.0
+ */
+@FunctionalInterface
+public interface HttpAnnotationProcessor {
+
+  /**
+   * Builds a preprocessor for the annotation which is supported by this processor. The preprocessor will be called
+   * before the actual http handler is called, allowing the parser to modify or event abort the request.
```
0utplay: even abort

driver/src/main/java/eu/cloudnetservice/driver/network/http/annotation/parser/HttpAnnotationProcessorUtil.java
```diff
@@ -0,0 +1,59 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.BiFunction;
+import lombok.NonNull;
+
+public final class HttpAnnotationProcessorUtil {
```
0utplay, line 34: No javadoc here?

0utplay (changes requested):
Looks great

0utplay (approved):

PR #699, opened 2022-07-02
chore: add support for compressed http responses

derklaro:
### Motivation
Clients can send an `accept-encoding` http header to request a specific encoding of the response, some of them require the server to compress the response (especially useful when trying to f. ex. download files).

### Modification
Add the content compressor handler of netty into the pipeline which automatically compresses the response when requested. (Supported compressions are `gzip`, `deflate`. `zstd` & `br`). 

### Result
We are now compressing http responses when requested.


schucko1337 (approved):

driver/src/test/java/eu/cloudnetservice/driver/network/http/NettyHttpServerTest.java
```diff
@@ -377,6 +379,48 @@ public void handle(String path, HttpContext context) {
   @Test
   @Order(90)
   @Timeout(20)
+  void testCompressionRequest() throws Exception {
+    var port = this.randomFreePort();
+    HttpServer server = new NettyHttpServer();
```
0utplay, line 384: Use var

0utplay (changes requested):

0utplay (approved):

PR #710, opened 2022-07-09
chore: migrate http handlers to annotation based api

derklaro:
### Motivation
We recently introduced a new http annotation api and all of our existing handlers should use that api as well as it cleans up a lot of code and is much simpler to understand.

### Modification
 - Migrate all existing http handlers to the annotation based api
 - Fix some request handling inconsistencies
 - Fix wrong `cancelNext` call in all handlers (without a paramater it's a getter not a setter)
 - Remove WebSocket ticket system as it has no benifit
 - Remove file serve http helpers to prevent users from building a file serving server using inbuild api
 - Fix some last issues discovered with the annotation api while migrating

### Result
Much cleaner http handlers and a lot less code overall.


schucko1337 (approved):

modules/rest/src/main/java/eu/cloudnetservice/modules/rest/v2/V2HttpHandlerCluster.java
```diff
@@ -196,36 +174,33 @@ protected void handleNodeUpdateRequest(@NonNull HttpContext context) {
...
       .append("nodeInfoSnapshot", node.nodeInfoSnapshot());
   }
 
-  protected @Nullable NodeServer nodeServer(@NonNull HttpContext context) {
-    var nodeName = context.request().pathParameters().get("node");
-    return this.nodeProvider().node(nodeName);
-  }
-
-  protected @NonNull NodeServerProvider nodeProvider() {
+  @NonNull
```
0utplay: Cant we inline the annotation

0utplay (changes requested):

modules/rest/src/main/java/eu/cloudnetservice/modules/rest/v2/V2HttpHandlerTemplate.java
```diff
@@ -264,74 +290,95 @@ protected void handleInstallationRequest(@NonNull HttpContext context) {
         .build();
 
       if (this.node().serviceVersionProvider().installServiceVersion(installer, forceInstall)) {
-        this.ok(context).body(this.success().toString()).context().closeAfter(true).cancelNext();
+        this.ok(context).body(this.success().toString()).context().closeAfter(true).cancelNext(true);
```
0utplay, line 293: Can we reformat this

0utplay (changes requested):

PR #712, opened 2022-07-09
[ci skip] chore: improve readme

derklaro:
### Motivation
The old readme file is very overloaded with information that is not needed for users which are visiting our GitHub page and we should focus more on developers visiting it.

### Modification
Cleanup and improve the readme in a way that it is much more focused on developers than users.

### Result
A better readme overall to present for repo visitors.


schucko1337 (approved):
Thank you

0utplay (approved):

PR #727, opened 2022-07-16
feat: Allow a host address on a per task basis

0utplay:
### Motivation
Currently every service is started on the same host address provided in the node configuration. The problem with this is that there is no way to use different hosts for different tasks like a Proxy and a Lobby.
### Modification
Created a new host address option for tasks which is used to bind the services of the task to. In order to make this feature available in the cluster too (different ips on different host systems), an ip alias system was implemented. The ip aliases are configured on each node and are resolved when starting the service. If somebody wants to use the same ip for each task supplying null as host address will use the fallback host, which is the old approach of using the host address provided by the node.

### Result
All tasks can have different host addresses and bind their services to that host address.

### DIFF
--- a/node/src/main/java/eu/cloudnetservice/node/command/sub/ClusterCommand.java
+++ b/node/src/main/java/eu/cloudnetservice/node/command/sub/ClusterCommand.java
@@ -43,8 +43,8 @@ import eu.cloudnetservice.node.command.annotation.Description;
 import eu.cloudnetservice.node.command.exception.ArgumentNotAvailableException;
 import eu.cloudnetservice.node.command.source.CommandSource;
 import eu.cloudnetservice.node.command.source.ConsoleCommandSource;
+import eu.cloudnetservice.node.util.NetworkUtil;
 import java.io.IOException;
-import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.text.DateFormat;
@@ -128,24 +128,46 @@ public final class ClusterCommand {
       .toList();
   }
 
-  @Parser
+  @Parser(name = "anyHostAndPort")
   public @NonNull HostAndPort defaultHostAndPortParser(@NonNull CommandContext<?> $, @NonNull Queue<String> input) {
     var address = input.remove();
-    try {
-      // create an uri with the tpc protocol
-      var uri = URI.create("tcp://" + address);
-
-      var host = uri.getHost();
-      // check if the host and port are valid
-      if (host == null || uri.getPort() == -1) {
-        throw new ArgumentNotAvailableException(I18n.trans("command-cluster-invalid-host"));
-      }
-
-      var inetAddress = InetAddresses.forUriString(host);
-      return new HostAndPort(inetAddress.getHostAddress(), uri.getPort());
-    } catch (IllegalArgumentException exception) {
-      throw new ArgumentNotAvailableException(I18n.trans("command-cluster-invalid-host"));
+    var hostAndPort = NetworkUtil.parseHostAndPort(address, true);
+    if (hostAndPort == null) {
+      throw new ArgumentNotAvailableException(I18n.trans("command-any-host-and-port-invalid", address));
     }
+
+    return hostAndPort;
+  }
+
+  @Parser(name = "assignableHostAndPort", suggestions = "assignableHostAndPort")
+  public @NonNull HostAndPort assignableHostAndPortParser(@NonNull CommandContext<?> $, @NonNull Queue<String> input) {
+    var address = input.remove();
+    var hostAndPort = NetworkUtil.parseHostAndPort(address, false);
+    // check if we can parse a host and port form the given input
+    if (hostAndPort == null) {
+      throw new ArgumentNotAvailableException(I18n.trans("command-any-host-invalid", address));
+    }
+    // check if we can assign the parsed host and port
+    if (NetworkUtil.assignableHostAndPort(hostAndPort) == null) {
+      throw new ArgumentNotAvailableException(I18n.trans("command-assignable-host-invalid", address));
+    }
+    // success
+    return hostAndPort;
+  }
+
+  @Suggestions("assignableHostAndPort")
+  public @NonNull List<String> suggestAssignableHostAndPort(@NonNull CommandContext<?> $, @NonNull String input) {
+    return List.copyOf(NetworkUtil.availableIPAddresses());
+  }
+
+  @Parser(name = "anyHost")
+  public @NonNull String anyHostParser(@NonNull CommandContext<?> $, @NonNull Queue<String> input) {
+    var address = input.remove();
+    if (InetAddresses.isInetAddress(address)) {
+      return address;
+    }
+
+    throw new ArgumentNotAvailableException(I18n.trans("command-any-host-invalid", address));
   }
 
   @Parser(name = "noNodeId", suggestions = "clusterNode")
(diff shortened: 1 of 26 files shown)


node/src/main/java/eu/cloudnetservice/node/util/NetworkUtil.java
```diff
@@ -68,6 +75,86 @@ public static boolean isInUse(int port) {
...
+    }
+  }
+
+  public static @Nullable HostAndPort assignableHostAndPort(@NonNull String address, boolean withPort) {
+    var hostAndPort = parseHostAndPort(address, withPort);
+    // could not parse host and port from string
+    if (hostAndPort == null) {
+      return null;
+    }
+    return assignableHostAndPort(hostAndPort);
```
derklaro:
```suggestion
    // try to parse host and port from the given string
    var hostAndPort = parseHostAndPort(address, withPort);
    return hostAndPort == null ? null : assignableHostAndPort(hostAndPort);
```

node/src/main/java/eu/cloudnetservice/node/util/NetworkUtil.java
```diff
@@ -68,6 +75,86 @@ public static boolean isInUse(int port) {
     }
   }
 
+  public static @Nullable HostAndPort assignableHostAndPort(@NonNull HostAndPort hostAndPort) {
+    try (var socket = new ServerSocket()) {
+      // try to bind on the given address
+      socket.bind(new InetSocketAddress(hostAndPort.host(), 45555));
```
derklaro:
```suggestion
      socket.bind(new InetSocketAddress(hostAndPort.host(), 0));
```
Should bind the socket to the next free port

node/src/main/java/eu/cloudnetservice/node/util/NetworkUtil.java
```diff
@@ -68,6 +75,86 @@ public static boolean isInUse(int port) {
     }
   }
 
+  public static @Nullable HostAndPort assignableHostAndPort(@NonNull HostAndPort hostAndPort) {
```
derklaro: This name seems off - shouldn't it be something like `checkAssignable`?

node/src/main/java/eu/cloudnetservice/node/util/NetworkUtil.java
```diff
@@ -68,6 +75,86 @@ public static boolean isInUse(int port) {
...
+      socket.bind(new InetSocketAddress(hostAndPort.host(), 45555));
+      return hostAndPort;
+    } catch (IOException exception) {
+      return exception instanceof BindException
+        && exception.getMessage() != null
+        && exception.getMessage().startsWith("Address already in use") ? hostAndPort : null;
+    }
+  }
+
+  public static @Nullable HostAndPort assignableHostAndPort(@NonNull String address, boolean withPort) {
```
derklaro: Same here, the name is off - it should be prefixed with `parse`

node/src/main/java/eu/cloudnetservice/node/command/sub/ClusterCommand.java
```diff
@@ -128,24 +128,46 @@ public final class ClusterCommand {
...
+
+  @Suggestions("assignableHostAndPort")
+  public @NonNull List<String> suggestAssignableHostAndPort(@NonNull CommandContext<?> $, @NonNull String input) {
+    return List.copyOf(NetworkUtil.availableIPAddresses());
+  }
+
+  @Parser(name = "anyHost")
+  public @NonNull String anyHostParser(@NonNull CommandContext<?> $, @NonNull Queue<String> input) {
+    var address = input.remove();
+    if (InetAddresses.isInetAddress(address)) {
```
derklaro: Hm, maybe we should use our parsing method here as well?

node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
```diff
@@ -163,6 +164,32 @@ public static void applyServiceConfigurationDisplay(
     return this.taskProvider().serviceTasks().stream().map(Nameable::name).toList();
   }
 
+  @Parser(suggestions = "ipAliasHostAddress", name = "ipAliasHostAddress")
+  public @NonNull String hostAddressParser(@NonNull CommandContext<?> $, @NonNull Queue<String> input) {
```
derklaro, line 168: A lot of spaces here 😂

node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
```diff
@@ -328,6 +355,21 @@ public void setMinServiceCount(
     }
   }
 
+  @CommandMethod("tasks task <name> set hostAddress <hostAddress>")
+  public void setHostAddress(
+    @NonNull CommandSource source,
+    @NonNull @Argument("name") Collection<ServiceTask> serviceTasks,
+    @NonNull @Argument(value = "hostAddress", parserName = "ipAliasHostAddress") String hostAddress
```
derklaro, line 362: But with this there is no way to reset the host address? (aside from editing the file)

node/src/main/java/eu/cloudnetservice/node/config/JsonConfiguration.java
```diff
@@ -58,6 +60,19 @@ public final class JsonConfiguration implements Configuration {
     return null;
   };
 
+  private static final Function<String, Map<String, String>> MAP_PARSER = value -> {
+    var values = value.split(";;");
+    Map<String, String> results = new HashMap<>();
+
+    for (var pair : values) {
```
derklaro: Can we move the split directly in the for loop?

node/src/main/java/eu/cloudnetservice/node/config/JsonConfiguration.java
```diff
@@ -466,6 +488,16 @@ public void connectHostAddress(@NonNull String connectHostAddress) {
     this.connectHostAddress = connectHostAddress;
   }
 
+  @Override
+  public @NonNull Map<String, String> ipAliases() {
+    return this.ipAliases != null ? this.ipAliases : (this.ipAliases = new HashMap<>());
```
derklaro: In theory the ipAliases map should always be present after the call to `load`?

node/src/main/java/eu/cloudnetservice/node/setup/SpecificTaskSetup.java
```diff
@@ -90,6 +92,21 @@ public void applyQuestions(@NonNull ConsoleSetupAnimation animation) {
...
           .recommendation(44955))
         .build(),
+      QuestionListEntry.<String>builder()
+        .key("taskHostAddress")
+        .translatedQuestion("command-tasks-setup-question-host-address")
+        .answerType(QuestionAnswerType.<String>builder()
+          .recommendation(Node.instance().config().hostAddress())
+          .parser(Parsers.assignableHostAndPortOrAlias())
+          .possibleResults(() ->
+            NetworkUtil.availableIPAddresses()
```
derklaro:
```suggestion
          .possibleResults(() -> NetworkUtil.availableIPAddresses()
```

node/src/main/java/eu/cloudnetservice/node/command/sub/ClusterCommand.java
```diff
@@ -128,24 +128,46 @@ public final class ClusterCommand {
...
+    if (NetworkUtil.assignableHostAndPort(hostAndPort) == null) {
+      throw new ArgumentNotAvailableException(I18n.trans("command-assignable-host-invalid", address));
     }
+    // success
+    return hostAndPort;
+  }
+
+  @Suggestions("assignableHostAndPort")
+  public @NonNull List<String> suggestAssignableHostAndPort(@NonNull CommandContext<?> $, @NonNull String input) {
+    return List.copyOf(NetworkUtil.availableIPAddresses());
```
derklaro, line 158: (Cannot comment on the method in NetworkUtil so doing it here): Shouldn't we move the `availableIPAddresses` into a static field which gets resolved only once? Especially if we have that many acesses to the list...

derklaro (changes requested):
Quick feedback

node/src/main/java/eu/cloudnetservice/node/util/NetworkUtil.java
derklaro: Any maybe return a boolean 🤔

derklaro (approved):

schucko1337 (approved):

PR #733, opened 2022-07-18
[ci skip] chore: update default service configuration files

derklaro:
### Motivation
Some configuration shipped with CloudNet are outdated and should be updated.

### Modification
Regenerated all configuration files and updated them in CloudNet. Some config values are now restored to the defaults as well.

### Result
Updated and actual default configs.


schucko1337 (approved):

node/src/main/resources/files/nms/server.properties
```diff
@@ -18,40 +18,38 @@ difficulty=easy
...
 white-list=false
 broadcast-console-to-ops=true
-previews-chat=true
 spawn-npcs=true
+previews-chat=false
 spawn-animals=true
-snooper-enabled=true
 function-permission-level=2
-level-type=default
+level-type=minecraft\:normal
```
0utplay, line 52: does this work in older versions like 1.8.8 too?

derklaro: sure

0utplay (approved):

PR #737, opened 2022-07-21
fix: allow setting commands to null

0utplay:
### Motivation
Currently there is no way to set null values in tasks using a command.

### Modification
Created some unset commands so that you can set null values using a command.
### Result
Null values can be set using commands.

### DIFF
--- a/node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
+++ b/node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
@@ -761,6 +761,48 @@ public final class TasksCommand {
     }
   }
 
+  @CommandMethod("tasks task <name> unset javaCommand")
+  public void unsetJavaCommand(
+    @NonNull CommandSource source,
+    @NonNull @Argument("name") Collection<ServiceTask> serviceTasks
+  ) {
+    for (var serviceTask : serviceTasks) {
+      this.updateTask(serviceTask, builder -> builder.javaCommand(null));
+      source.sendMessage(I18n.trans("command-tasks-set-property-success",
+        "javaCommand",
+        serviceTask.name(),
+        "null"));
+    }
+  }
+
+  @CommandMethod("tasks task <name> unset hostAddress")
+  public void unsetHostAddress(
+    @NonNull CommandSource source,
+    @NonNull @Argument("name") Collection<ServiceTask> serviceTasks
+  ) {
+    for (var serviceTask : serviceTasks) {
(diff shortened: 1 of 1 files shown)


node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
```diff
@@ -761,6 +761,48 @@ public void clearProcessParameter(
...
+    }
+  }
+
+  @CommandMethod("tasks task <name> unset runtime")
+  public void unsetRuntime(
+    @NonNull CommandSource source,
+    @NonNull @Argument("name") Collection<ServiceTask> serviceTasks
+  ) {
+    for (var serviceTask : serviceTasks) {
+      this.updateTask(serviceTask, builder -> builder.runtime(null));
```
derklaro: which runtime is used when we set it to null?

node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
```diff
@@ -761,6 +761,48 @@ public void clearProcessParameter(
...
   }
 
+  @CommandMethod("tasks task <name> unset javaCommand")
+  public void unsetJavaCommand(
+    @NonNull CommandSource source,
+    @NonNull @Argument("name") Collection<ServiceTask> serviceTasks
+  ) {
+    for (var serviceTask : serviceTasks) {
+      this.updateTask(serviceTask, builder -> builder.javaCommand(null));
+      source.sendMessage(I18n.trans("command-tasks-set-property-success",
```
schucko1337, line 771: I dont like this playo. This looks like a lot of duplicate code :( but if klaro likes it then it's okay.

0utplay: How would you like to do it?

node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
derklaro: this is not resolved

node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
derklaro: I think it's about moving the for loop into a seperate method, and I can see the issue. Probably best to move into another pull request tho? @robertschuck

derklaro (approved):

PR #754, opened 2022-08-04
Added minimessage support for syncproxy

Antimoxs:


wasn325:
You should maybe check, if the license header is applied everywhere

PR #760, opened 2022-08-06
chore: use virtual threads for network processing if possible

derklaro:
### Motivation
JDK 19 added the virtual thread feature as a preview feature. As these "threads" are not backed by a platform thread, and are even sleeping while waiting for some operations (like file system aceess) they are the perfect fit for our packet processing work with can require a lot of threads which are sometimes just waiting for background work to complete.

### Modification
Added a util class which can detect when runnning on Java 19+ with preview features enabled, and build virtual thread executors when needed. These are for now only used for packet dispatching. If virtual threads are not available, then the caller can specify a fallback thread factory to use.

### Result
Packet dispatching is optimized for users which are using newer jvms with preview features enabled.

##### Other context
Some nice stuff to read:
 * https://openjdk.org/jeps/425
 * https://blogs.oracle.com/javamagazine/post/java-loom-virtual-threads-platform-threads

### DIFF
--- a/driver/src/main/java/eu/cloudnetservice/driver/network/netty/NettyUtil.java
+++ b/driver/src/main/java/eu/cloudnetservice/driver/network/netty/NettyUtil.java
@@ -16,10 +16,10 @@
 
 package eu.cloudnetservice.driver.network.netty;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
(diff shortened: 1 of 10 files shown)


0utplay (approved):

schucko1337 (approved):
nice. thank you

PR #802, opened 2022-08-19
fix: correctly resolve dependencies for cnl files during compile

derklaro:
### Motivation
The current dependency resolution system tries to resolve the internal modules (like driver, common etc) from the central repository as they are available there now. This leads to errors during startup, as well as the system isn't reporting failed repository lookups which are only visible in runtime.

### Modification
Exclude all dependencies with the cloudnet group id and throw an exception if the repo resolution fails.

### Result
Repository resolution and dependency inclusion now works as expected again.


schucko1337 (approved):
ok. this is the last time I approve kotlin Prs.

PR #817, opened 2022-08-26
chore: improve cloudflare module

derklaro:
### Motivation
The current CloudFlare module implementation isn't done as well as it could be.

### Modification
Change a lot of stuff of the module internals, but also expose many new stuff which will improve the quality for networks drastically:
 * configuration entries now have a name setting which allows to modify the subdomain of the node entry which will be created. That is especially useful considering that now proxies can be bound to different addresses, but the CloudFlare module was only creating one entry per zone to point all proxies to.
 * Node records are now only deleted on stop when the configuration entry from which they were created originally no longer exists. Entries which get deleted while the module is running will be there until the module stops to prevent issues with other records depending on the entry.
 * Changes of configuration entries are now applied in realtime when reloading the module. For example adding a new entry will create the correct A/AAAA record, and changing an entry will patch the record as needed. Comparison of entries are done based on the zone id & entry name (the main 2 identification points which are required to be fix).

### Result
A much better implementation of the CloudFlare module.


modules/cloudflare/src/main/java/eu/cloudnetservice/modules/cloudflare/CloudNetCloudflareModule.java
```diff
@@ -54,78 +66,178 @@ public void convertConfiguration() {
...
+    // find all newly added records - while we could also remove all old records, this could result in unexpected issues
+    // like proxy SRV records pointing to a deleted record
+    var stillExistingEntries = newEntries.stream()
+      .filter(CloudflareConfigurationEntry::enabled)
+      .flatMap(entry -> oldEntries.stream()
+        .filter(old -> CloudflareConfigurationEntry.mightEqual(old, entry))
+        .map(oldEntry -> new Pair<>(entry, oldEntry)))
+      .toList();
+
+    // filter out differing entries which still exists, without including the group-based configuration
```
0utplay:
```suggestion
    // filter out differing entries which still exist, without including the group-based configuration
```

modules/cloudflare/src/main/java/eu/cloudnetservice/modules/cloudflare/CloudNetCloudflareModule.java
```diff
@@ -54,78 +66,178 @@ public void convertConfiguration() {
...
+
+        // always delete service records
+        return true;
+      })
+      .map(Map.Entry::getValue)
+      .map(this.recordManager::deleteRecord)
+      .toArray(CompletableFuture[]::new);
+
+    // create one final future from all deletion futures and wait
+    // for them to complete or time them out after 10 seconds
```
0utplay:
```suggestion
    // for them to complete or time them out after 15 seconds
```

0utplay (changes requested):

PR #829, opened 2022-08-30
chore: improve logging when launcher update repo/branch wasn't found

derklaro:
### Motivation
Currently when we get an invalid status (other than 200) from the remote server we're just throwing an exception which gives closes to 0 information to the end user and the support.

### Modification
Remove throwing the exception when an invalid status gets send by the remote server and replace it with proper logging.

### Result
Better errors for the user and hints for the support.


0xIO32:
I think for the self-fixing is a stacktrace more useful than a logged error message. Then else you know only the error but not from where the error is thrown.
Or a DEBUG-Message with the stack trace -> Thread.currentThread().getStackTrace();

derklaro:
No, it's very clear in that case what the issue is - there is no need for a stacktrace

0utplay (approved):

PR #853, opened 2022-09-14
Fixed server kick

davidmayr:
### Motivation
When a server kicks a player while logging in and the player wants to join again the player will get kicked from the network stating that no lobby server could be found.

### Modification

In Velocity I moved the notify result code so it runs before fallbacks are getting queried. In Bungeecord I tried to implement the same functionallity altough I haven't tested that yet and don't have much experience with the Bungeecord-API. 

### Result

This PR stops the player from getting kicked from the entire network as long as the player has an active connection to a server.

### DIFF
--- a/modules/bridge/src/main/java/eu/cloudnetservice/modules/bridge/platform/bungeecord/BungeeCordPlayerManagementListener.java
+++ b/modules/bridge/src/main/java/eu/cloudnetservice/modules/bridge/platform/bungeecord/BungeeCordPlayerManagementListener.java
@@ -118,12 +118,16 @@ public final class BungeeCordPlayerManagementListener implements Listener {
   @EventHandler(priority = EventPriority.LOWEST)
   public void handle(@NonNull ServerKickEvent event) {
     if (event.getPlayer().isConnected()) {
-      if (event.getState() == ServerKickEvent.State.CONNECTING && event.getPlayer().getServer() != null) {
-        event.setCancelServer(event.getPlayer().getServer().getInfo());
+      // only notify the player if the player was kicked on login and is already connected to a server
+      var curServer = event.getPlayer().getServer();
+
+      if (event.getState() == ServerKickEvent.State.CONNECTING && curServer != null) {
+        event.setCancelServer(curServer.getInfo());
         event.setCancelled(true);
         this.sendKickMessage(event);
         return;
       }
+
       ServerInfo target = this.management.fallback(event.getPlayer(), event.getKickedFrom().getName())
         .map(service -> ProxyServer.getInstance().getServerInfo(service.name()))
         .orElse(null);
@@ -146,7 +150,7 @@ public final class BungeeCordPlayerManagementListener implements Listener {
 
   private void sendKickMessage(@NonNull ServerKickEvent event) {
     // extract the reason for the disconnect and wrap it
-    Locale playerLocale = event.getPlayer().getLocale();
+    var playerLocale = event.getPlayer().getLocale();
     var baseMessage = this.management.configuration().message(playerLocale, "error-connecting-to-server")
       .replace("%server%", event.getKickedFrom().getName())
       .replace("%reason%", BaseComponent.toLegacyText(event.getKickReasonComponent()));
--- a/modules/bridge/src/main/java/eu/cloudnetservice/modules/bridge/platform/bungeecord/BungeeCordPlayerManagementListener.java
+++ b/modules/bridge/src/main/java/eu/cloudnetservice/modules/bridge/platform/bungeecord/BungeeCordPlayerManagementListener.java
@@ -43,7 +43,6 @@ import net.md_5.bungee.api.plugin.Listener;
 import net.md_5.bungee.api.plugin.Plugin;
 import net.md_5.bungee.event.EventHandler;
 import net.md_5.bungee.event.EventPriority;
-import org.jetbrains.annotations.NotNull;
 
 public final class BungeeCordPlayerManagementListener implements Listener {
 
@@ -122,7 +121,7 @@ public final class BungeeCordPlayerManagementListener implements Listener {
       if (event.getState() == ServerKickEvent.State.CONNECTING && event.getPlayer().getServer() != null) {
         event.setCancelServer(event.getPlayer().getServer().getInfo());
         event.setCancelled(true);
-        sendKickMessage(event);
+        this.sendKickMessage(event);
         return;
       }
       ServerInfo target = this.management.fallback(event.getPlayer(), event.getKickedFrom().getName())
@@ -133,7 +132,7 @@ public final class BungeeCordPlayerManagementListener implements Listener {
       if (target != null) {
         event.setCancelled(true);
         event.setCancelServer(target);
-        sendKickMessage(event);
+        this.sendKickMessage(event);
       } else {
         // no lobby server - the player will disconnect
         event.setCancelled(false);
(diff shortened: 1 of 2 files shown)


modules/bridge/src/main/java/eu/cloudnetservice/modules/bridge/platform/bungeecord/BungeeCordPlayerManagementListener.java
```diff
@@ -118,20 +118,21 @@ public void handle(@NonNull ServerConnectEvent event) {
   @EventHandler(priority = EventPriority.LOWEST)
   public void handle(@NonNull ServerKickEvent event) {
     if (event.getPlayer().isConnected()) {
+      if (event.getState() == ServerKickEvent.State.CONNECTING && event.getPlayer().getServer() != null) {
```
derklaro: Please keep the code consistent with the velocity one:
 * Extract the current server into a variable
 * Please add a comment why we're doing this
 * Please add a blank line after the if block

modules/bridge/src/main/java/eu/cloudnetservice/modules/bridge/platform/bungeecord/BungeeCordPlayerManagementListener.java
```diff
@@ -143,6 +144,16 @@ public void handle(@NonNull ServerKickEvent event) {
     }
   }
 
+  private void sendKickMessage(@NonNull ServerKickEvent event) {
+    // extract the reason for the disconnect and wrap it
+    Locale playerLocale = event.getPlayer().getLocale();
```
derklaro:
```suggestion
    var playerLocale = event.getPlayer().getLocale();
```

modules/bridge/src/main/java/eu/cloudnetservice/modules/bridge/platform/velocity/VelocityPlayerManagementListener.java
```diff
@@ -97,23 +96,24 @@ public void handleInitialServerChoose(@NonNull PlayerChooseInitialServerEvent ev
...
-        .map(server -> {
-          // only notify the player if the fallback is the server the player is connected to
-          var curServer = event.getPlayer().getCurrentServer().map(ServerConnection::getServerInfo).orElse(null);
-          if (event.kickedDuringServerConnect() && curServer != null && curServer.equals(server.getServerInfo())) {
-            // send the player a nice message - velocity will keep the connection to the current server
-            return KickedFromServerEvent.Notify.create(this.extractReasonComponent(event));
-          } else {
+      var curServer = event.getPlayer().getCurrentServer().orElse(null);
+
+      // only notify the player if the player was kicked on login and is already connected to a server
```
derklaro:
```suggestion
      // only notify the player if the player was kicked on login and is already connected to a server
      var curServer = event.getPlayer().getCurrentServer().orElse(null);
```

derklaro (changes requested):

derklaro:
Although this introduces another issue as well. If you're getting kicked during the connect to a fallback server, there are no further attempts to connect you to any other fallback service. The current code only sends the message if you're already connected to a fallback service. So I would prefer clearing the fallback profile of the player before the notify rather than changing the code in this way.

davidmayr:
> Although this introduces another issue as well. If you're getting kicked during the connect to a fallback server, there are no further attempts to connect you to any other fallback service. The current code only sends the message if you're already connected to a fallback service. So I would prefer clearing the fallback profile of the player before the notify rather than changing the code in this way.

Well I think it depends. When you are doing /hub and maybe have an issue connecting I see why this is an issue. But is it really a good idea sending the player to a new fallback when any login connection error occurs? I don't think the proxies do that by default and I see no reason it should behave that way(well except in the /hub case)

derklaro:
I'm always pro improving stuff, but this seems to (at the current point) introduce even more issues 🤔

davidmayr:
> I'm always pro improving stuff, but this seems to (at the current point) introduce even more issues 🤔

Is there any other issue then the /hub command only connecting to one server? Because if thats the only issue this stuff could be implemented in the hub command directly

0utplay:
Could you try using the change @derklaro made?

`fix-invalid-fallback-profile-entries` is the branch that contains the change.

davidmayr:
> Could you try using the change @derklaro made?
> 
> `fix-invalid-fallback-profile-entries` is the branch that contains the change.

It does work, but with more than one fallback running, you get thrown around from one fallback to another. On my branch you would remain on the same one, which is better to look at and saves some performance because the login process is pretty demanding

derklaro:
I don't get what you're testing, but based on my fix I don't understand how you would get thrown around...

davidmayr:
> I don't get what you're testing, but based on my fix I don't understand how you would get thrown around...

With one Lobby/fallback server there, your fix behaves exactly like mine. 
And the original issue is fixed on your branch too, but the behavior still annoys me a bit.

When there are two fallback servers e.g. Lobby-1 and Lobby-2 and you want to connect to a game server, but the game server doesn't let you in for whatever reason.(e.g. it's full) If you are on Lobby-1 you will get sent to Lobby-2 after the connection error occurred. If you are on Lobby-2 you will be sent to Lobby-1. In my opinion, these are just unnecessary connections that are happening since you are already connected to a fallback. 

Or let's say you are on a Hypixel Skyblock like Server and want to connect to your Island (which is on a different sub-server) and a connection error occurs. Instead of remaining in the skyblock hub and receiving an error, you will end up in the main hub. I just don't like that behavior.

derklaro:
> If you are on Lobby-2 you will be sent to Lobby-1. In my opinion, these are just unnecessary connections that are happening since you are already connected to a fallback.

Correct, but it still seems to me that your fix is adding an unexpected behaviour if a fallback service is rejecting a player connection. Consider the following: A BedWars game ended, the server is kicking all the players, the lobby server is rejecting the connection (for example because it's full) & the players stays on the BedWars service. The connection to the server times out but there are no attemps made to connect the player to a different fallback.

> Or let's say you are on a Hypixel Skyblock like Server and want to connect to your Island (which is on a different sub-server) and a connection error occurs. Instead of remaining in the skyblock hub and receiving an error, you will end up in the main hub. I just don't like that behavior.

That is wrong & fixed with my proposed fix.

davidmayr:
> A BedWars game ended, the server is kicking all the players, the lobby server is rejecting the connection (for example because it's full) & the players stays on the BedWars service

But wouldn't the connection be closed, and the kick event code is forced to search for a new one? Because my changes would only apply if the player has an active connection. 


> That is wrong & fixed with my proposed fix.

It isn't. If you are on SubServer-1 and want to connect to SubServer-2, but an error occurs, you will be connected to a fallback Lobby. And in my SkyBlock example, this would be the same scenario.

0utplay:
I just tested this and there is now way to leave a service using /hub if the first fallback does not accept the request (but there is another that could accept it). You are basically stuck on the service.

davidmayr:
> I just tested this and there is now way to leave a service using /hub if the first fallback does not accept the request (but there is another that could accept it). You are basically stuck on the service.

That is solvable, though, by implementing/modifying the callback in the hub command and making it search for more servers.

derklaro:
But this would require every plugin to adopt to your change, and every plugin to manually maintain the tried fallbacks of the player which can be a pitfall if you're not doing that carefully. I don't think that in the current state we can/should accept this pull request.

PR #865, opened 2022-09-25
feat(npcs): add a command to remove all spawned npcs

0utplay:
### Motivation
Currently there is no way to bulk delete npcs, you have to visit each npc and delete it with the remove command once at a time.

### Modification
Added a new `removeall` command which allows to delete all npcs or all npcs that target a specific group.

### Result
Bulk deletion of npcs is possible using the mentioned command.

### DIFF
--- a/modules/npcs/src/main/java/eu/cloudnetservice/modules/npc/platform/bukkit/command/NPCCommand.java
+++ b/modules/npcs/src/main/java/eu/cloudnetservice/modules/npc/platform/bukkit/command/NPCCommand.java
@@ -151,6 +151,20 @@ public final class NPCCommand extends BaseTabExecutor {
       return true;
     }
 
+    // npc remove all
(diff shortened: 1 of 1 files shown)


modules/npcs/src/main/java/eu/cloudnetservice/modules/npc/platform/bukkit/command/NPCCommand.java
```diff
@@ -151,6 +151,20 @@ public boolean onCommand(CommandSender sender, Command command, String label, St
       return true;
     }
 
+    // npc remove all
+    if (args.length >= 1 && args[0].equalsIgnoreCase("removeall")) {
+      int removedNpcs;
+
+      if (args.length == 2) {
+        removedNpcs = this.management.deleteAllNPCs(args[1]);
```
derklaro, line 160: Please add a comment what `args[1]` is

derklaro (approved):

PR #879, opened 2022-10-01
feat: allow setting environment variables for services

derklaro:
### Motivation
We currently only allow users and developers to configure jvm options and process paramaters, but exclude the possiblity to set environment variables. As some server implementations (like paper) allow setting some configuration options using environment variables, this option might become handy.

### Modification
Add the possibility to set environment variables in tasks, groups and service configurations. These variables are parsed during the start of a service. Keys of environment variables are always converted to uppercase, as docker rejects variables with a lowercase name.

### Result
Users and developers are now able to set environment variables for services.


node/src/main/java/eu/cloudnetservice/node/provider/NodeGroupConfigurationProvider.java
```diff
@@ -191,15 +192,24 @@ protected void writeAllGroupConfigurations() {
 
   protected void loadGroupConfigurations() {
     FileUtil.walkFileTree(GROUP_DIRECTORY_PATH, ($, file) -> {
+      var document = JsonDocument.newDocument(file);
+
+      // TODO: remove in 4.1
+      // check if the task has environment variables
```
0utplay, line 198: We are talking about groups here

0utplay (changes requested):

node/src/main/java/eu/cloudnetservice/node/provider/NodeGroupConfigurationProvider.java
derklaro: Not really relevant, we're removing these lines in a few versions anyway

0utplay (approved):

PR #880, opened 2022-10-02
fix: the template installer now installs the templates correctly

0utplay:
### Motivation
The template installer of the smart module was not working correctly as we did not have any information about templates that were really installed or just provided in the configuration.

### Modification
Added new installed templates, deployments and inclusions collections that track all templates etc. that were installed onto the service. Using these installed templates the template installer of the smart module now works like intended.

### Result
The template installer works correctly.

### DIFF
--- a/node/src/main/java/eu/cloudnetservice/node/event/service/CloudServicePostPrepareEvent.java
+++ b/node/src/main/java/eu/cloudnetservice/node/event/service/CloudServicePostPrepareEvent.java
@@ -19,7 +19,7 @@ package eu.cloudnetservice.node.event.service;
 import eu.cloudnetservice.node.service.CloudService;
 import lombok.NonNull;
 
-public class CloudServicePostPrepareEvent extends CloudServiceEvent {
+public final class CloudServicePostPrepareEvent extends CloudServiceEvent {
 
   public CloudServicePostPrepareEvent(@NonNull CloudService service) {
     super(service);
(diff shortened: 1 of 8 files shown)


node/src/main/java/eu/cloudnetservice/node/event/service/CloudServicePostPrepareEvent.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package eu.cloudnetservice.node.event.service;
+
+import eu.cloudnetservice.node.service.CloudService;
+import lombok.NonNull;
+
+public class CloudServicePostPrepareEvent extends CloudServiceEvent {
```
derklaro:
```suggestion
public final class CloudServicePostPrepareEvent extends CloudServiceEvent {
```

node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
```diff
@@ -636,10 +671,14 @@ protected void prepareService() {
     if (sslConfiguration.enabled()) {
       sslConfiguration = this.prepareSslConfiguration(sslConfiguration);
     }
+
     // add all components
     this.waitingTemplates.addAll(this.serviceConfiguration.templates());
     this.waitingDeployments.addAll(this.serviceConfiguration.deployments());
     this.waitingRemoteInclusions.addAll(this.serviceConfiguration.inclusions());
+
+    this.eventManager.callEvent(new CloudServicePrePrepareEvent(this));
```
derklaro, line 681:
```suggestion
    // initial service details are now ready, let the modules know that we're starting to prepare
    this.eventManager.callEvent(new CloudServicePrePrepareEvent(this));
```

derklaro (approved):
two small nitpicks

PR #920, opened 2022-10-15
chore: ensure that we can read a VarInt when decoding the packet length

derklaro:
### Motivation
There were recent reports from users which were getting exceptions caused by the packet decoder when reading a VarInt from the current buffer position. This might for example happen, if netty reads a chunk of data from a packet, then calls the deserializer but the buffer only contains one full packet and the first one - four bytes of the next packet (depending on the length). In that case, we're unable to read the next packet length and will fail with an exception. This will cause other packets to cause trouble as well, as the packet length will be read from a wrong position, causing the wrong length result which results in exceptions further down the line.

### Modification
Two changes were made to fix the issue and make the reading more stable in general:
* Ensure that we can read a VarInt from the buffer position when getting the packet. If we're unable to read a VarInt we shouldn't throw an exception but reset the reader index and wait for more data to arrive.
* Dump the packet hex and the full stack trace when trying to read a VarInt from a buffer. This helps to understand which data was send and where the read call came from which makes debugging much easier.

### Result
Hopefully no more exceptions caused by packet decoders because of an invalid packet length.


driver/src/main/java/eu/cloudnetservice/driver/network/netty/NettyUtil.java
```diff
@@ -156,10 +155,31 @@ private NettyUtil() {
...
+  }
+
+  /**
+   * Reads a var int from the given buffer, returns null if there is no Var Int at the current buffer position.
+   *
+   * @param buffer the buffer to read from.
+   * @return the var int read from the buffer, or null if no var int is at the given position.
+   * @throws NullPointerException if the given buffer to read from is null.
+   */
+  public static Integer readVarIntOrNull(@NonNull Buffer buffer) {
```
0utplay: @Nullable please

0utplay (approved):

PR #927, opened 2022-10-17
Chat color config change

### LINKED ISSUE #926: CloudNet-Chat ignores group color
### Stacktrace

_No response_

### Actions to reproduce

Version: CloudNet Blizzard 4.0.0-RC4 ab12241
Using both CloudNet-Chat and CloudNet-SimpleNameTags:
```
/cloud perms group admin set prefix "&x&f&f&0&0&f&fAdmin &4| "
/cloud perms group admin set color &7
```
Join the admin group.

Player list looks like this:
image

While chat looks like this, ignoring the color:
image

### CloudNet version

CloudNet Blizzard 4.0.0-RC4 ab12241

### Other

_No response_

### Issue uniqueness

Remynfv:
### Motivation

See issue #926 . Default tab list formatting with SimpleNameTags doesn't match the default CloudNet-Chat formatting.

### Modification

Changed the config.yml and config.properties to add color.

### Result

Chat is now formatting the same as the player list by default.

##### Other context

Fixes #926 

My first pull request to CloudNet! Hopefully I did this correctly.

### DIFF
--- a/plugins/chat/src/main/resources/config.properties
+++ b/plugins/chat/src/main/resources/config.properties
@@ -1,13 +1,13 @@
 # Available placeholders:
 # %group%    - the name of the highest permission group of the player
-# %display%  - the display of the highest permission group of the player
 # %prefix%   - the prefix of the highest permission group of the player
 # %suffix%   - the suffix of the highest permission group of the player
 # %color%    - the color of the highest permission group of the player
-# %name%     - the name of the player
+# %display%  - the display name of the player
+# %name%     - the username of the player
 # %uniqueId% - the unique id of the player
 # %message%  - the message sent by the player
 
 # permission to use "&" in messages for color: cloudnet.chat.color
 
-format=%display%%color%%name% &8:&f %message%
+format=%display%%color%%name%%suffix% &8:&f %message%
--- a/plugins/chat/src/main/resources/config.properties
+++ b/plugins/chat/src/main/resources/config.properties
@@ -10,4 +10,4 @@
 
 # permission to use "&" in messages for color: cloudnet.chat.color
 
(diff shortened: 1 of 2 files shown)


0utplay:
I am not sure if we should continue using the display as default here. We might consider switching to %prefix%%color%%name%

Remynfv:
%prefix% sounds like it would make more sense, but... what does %display% actually do?

> ``# %display%  - the display of the highest permission group of the player``

Maybe this could be more descriptive?

Also while we're at it, wouldn't it make sense to also use %suffix% at the end?

0utplay:
We could use %suffix% too, yeah. The display is set as the players display name (that does not appear in the TabList, but when using Player#getDisplayName()).

Remynfv:
setDisplayName is used by a lot of plugins to do nicknames and stuff. I feel like it could cause confusion if that didn't work out of the box. To me at least, _not_ using display names seems like the non-default behavior.

Remynfv:
I think it makes sense to keep %display%, especially for parity with SimpleNameTags. Simplenametags' does specifically mention display names as well.
>  Adds prefix, suffix and **display name** support to all server platforms

In these commits I also moved %display% to be next to %name%, and rephrased their descriptions to be a bit clearer.

0utplay (changes requested):
Hey, I would ask you to rebase your PR due to recent changes to the sponge impl.

Furtheremore please revert the comment change about the %display% option as it is indeed the display of the group. You can add %display_name% to the available placeholders list, which is the display name of the player.

Lastly, change the default formats in the Nukkit & Sponge code.

0utplay:
Closed due to inactivity

PR #938, opened 2022-10-24
fix: npcs should hide ingame services but still update the online count

### LINKED ISSUE #937: CloudNPC (Not updating)
### Actions to reproduce

If i wanna update the NPC MOTD or the MaxPlayers, it won't work. But if i respawn (edit) the NPC then it shows the right MOTD and player count. So basically the NPC won't update until i edit it.
image

### CloudNet version

 CloudNet Blizzard 4.0.0-RC4-ab12241b

### Other

_No response_

### Issue uniqueness

0utplay:
### Motivation
Currently npcs would not remove services that are in the INGAME state and also would not update their online count.
The online count / max players in the infoline would not match the real online count / max players.

### Modification
Removed the item from the inventory if the service is ingame and updated the info line.

### Result
The items are removed and updated correctly

##### Other context
Fixes #937


strich3:
Also, if we are already on that topic. Is there a way to display full services or ingame services in the inventory?

derklaro (approved):

PR #958, opened 2022-11-03
chore: improve console tab suggestions

derklaro:
### Motivation
The current tab completion process only takes arguments into account which exactly match the given input, allowing no error tolerance when typing. We can improve the tab completion to include results which match the given input to a certain degree.

### Modification
Made the tab completion compare results using the jaro-winkler string distance while requiring a minimum match of 0.7 (70%). This change also requires a small change to the console completion candidate processing to just accept previously filtered input rather then filtering it again.
The cloud fork version has been bumped as well to include a fix for invalid suggestions when a dynamically parsed argument wasn't parseable with the current given input.

### Result
The new console suggestions are less dependant on correctly typing an input and now contain results even if the spelling is a bit off. While the new tab completion might feel a bit off at first, this change should result in better completions & better, stable sorting of results.


node/src/main/java/eu/cloudnetservice/node/console/JLine3CompletionMatcher.java
```diff
@@ -0,0 +1,91 @@
...
+
+  @Override
+  public @Nullable Candidate exactMatch() {
+    // keep a local copy of the variables in case of concurrent calls
+    var candidates = Objects.requireNonNull(this.candidates);
+    var parsedLine = Objects.requireNonNull(this.parsedLine);
+
+    // check if there is a 100% match
+    var givenWord = parsedLine.word();
+    for (Candidate candidate : candidates) {
```
0utplay: var

node/src/main/java/eu/cloudnetservice/node/console/JLine3CompletionMatcher.java
```diff
@@ -0,0 +1,91 @@
...
+    return null;
+  }
+
+  @Override
+  public @Nullable String getCommonPrefix() {
+    // keep a local copy of the candidates in case of concurrent calls
+    var candidates = Objects.requireNonNull(this.candidates);
+
+    String commonPrefix = null;
+    for (Candidate candidate : candidates) {
```
0utplay: var

0utplay (approved):

PR #1059, opened 2023-01-09
[ci skip] chore: update copyright for 2023

drolpi:
### Motivation
Since the year 2023 has now begun, the year 2022, which is currently specified in the copyright header, is no longer current.

### Modification
Updates the copyright year in all copyright headers in all files to 2023 (included LICENSE_HEADER).

### Result
The copyright year in the license header is up to date again.

### DIFF
--- a/.template/start.command
+++ b/.template/start.command
@@ -1,19 +1,4 @@
 #!/bin/bash
-#
-# Copyright 2019-2023 CloudNetService team & contributors
-#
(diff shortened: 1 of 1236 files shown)


drolpi:
Review @derklaro @0utplay ?

derklaro:
Sorry for the late reply, but we do not accept changes that affect all files contributed by non-team members.

PR #1099, opened 2023-01-29
[ci skip] chore: rename the cloud service post construct listener

0utplay:
### Motivation
Every method should have a meaningful name.

### Modification
Changed the name of the post construct listener to a meaningful name.

### Result
The post construct listener has a meaningful name.


node/src/main/java/eu/cloudnetservice/node/service/defaults/DefaultCloudServiceManager.java
```diff
@@ -162,7 +162,7 @@ public DefaultCloudServiceManager(
   }
 
   @PostConstruct
-  private void a() {
+  private void registerDefaultServiceFactory() {
     // register the default factory
```
derklaro:
```suggestion
```

derklaro (approved):

PR #1163, opened 2023-03-14
fix: allow repetition of jvm options and command line arguments

0xIO32:
### Motivation
On my server I wanted to add the following JVM Options: "--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED", but noticed that duplicated elements e.g. "--add-opens" were removed automatically and thus this was not possible.

### Modification
I've modified the eu.cloudnetservice.driver.service.ProcessConfiguration Class to use a LinkedList instead of a LinkedHashSet

### Result
Now duplicated Elements in the jvmOptions and processParameters will now accept duplicated elements.


derklaro (approved):

0utplay (approved):

0utplay:
Thanks!

PR #1178, opened 2023-04-09
chore: rewrite document api

derklaro:
### Motivation
At the moment the document api is very limited due to the reason that practically only the json implementation is used everywhere. While there is an abstract interface for a Document  that specifies most of the methods implemented in the json implementation, that interface cannot really be used for custom implementations as the exposed api never accepts the custom implementation.

### Modification
The whole Document api has been rewritten and the Document interface is now used everywhere in the api, allowing to use custom document types. To make this work, this PR also contains a way to transfer the key-value pairs from one document type to another. To follow the internal structure of having one mutable type and one immutable type for everything (to make it clear when something can/should be edited, an when that shouldn't be the case) the document is now split up into mutable and immutable as well.

### Result
Way cleaner document implementation with proper support for custom document types.


driver/src/main/java/eu/cloudnetservice/driver/document/Document.java
```diff
@@ -0,0 +1,854 @@
...
+ * Due to the fact that a document is not representing a specific implementation (for example JSON) there is no
+ * guarantee that an implementation supports all methods to read elements from the underlying configuration type. In
+ * these cases the called methods will throw an {@link UnsupportedOperationException}. If a conversion between different
+ * document types is required, use the send api as follows:
+ * <pre>
+ * {@code
+ *   Document jsonDocument = Document.newJsonDocument().append("json_key", "Hello World!");
+ *   Document.Mutable tomlDocument = TomlDocumentSource.newDocument().append("toml_key", "CloudNet");
+ *
+ *   // Serializes the content of the json document into a format than
```
0utplay:
```suggestion
 *   // Serializes the content of the json document into a format that
```

driver/src/main/java/eu/cloudnetservice/driver/document/DocumentSerialisationException.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.Serial;
+import lombok.NonNull;
+
+/**
+ * An exception that gets thrown when an issue occurs while serializing a document.
+ *
+ * @since 4.0
+ */
+public final class DocumentSerialisationException extends RuntimeException {
```
0utplay, line 27:
```suggestion
public final class DocumentSerializationException extends RuntimeException {
```

driver/src/main/java/eu/cloudnetservice/driver/module/driver/DriverModule.java
```diff
@@ -50,22 +51,21 @@ public class DriverModule extends DefaultModule {
...
    * into a json document, throwing an exception when the document is invalid.
    *
+   * @param factory the config factory to use when parsing the configuration.
    * @return the config at the default path.
    * @throws JsonSyntaxException if the document is invalid.
-   * @see JsonDocument#newDocument(Path)
    */
-  public @NonNull JsonDocument readConfig() {
-    return JsonDocument.newDocument(this.configPath());
+  public @NonNull Document readConfig(@NonNull DocumentFactory factory) {
```
0utplay, line 58: Maybe we should consider adding a readJsonConfig method as short cut?

0utplay (approved):

driver/src/main/java/eu/cloudnetservice/driver/document/DocumentSerialisationException.java
derklaro: no

driver/src/main/java/eu/cloudnetservice/driver/module/driver/DriverModule.java
derklaro: not necessary imo

driver/src/main/java/eu/cloudnetservice/driver/document/Document.java
derklaro: fixed

PR #1188, opened 2023-04-11
fix: headers in service inclusions should be optional

0utplay:
### Motivation
When including an inclusion that has no headers option set the following exception is thrown.
https://cdn.discordapp.com/attachments/818777626663321671/1095292788565868645/image.png

### Modification
Made sure that we read an empty document if the headers option is not set.

### Result
The header option is now optional as documented.


node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
```diff
@@ -359,7 +359,7 @@ public void includeWaitingServiceInclusions() {
       // prepare the connection from which we load the inclusion
       var req = Unirest.get(inclusion.url());
       // put the given http headers
-      var headers = inclusion.readProperty(ServiceRemoteInclusion.HEADERS);
+      var headers = inclusion.readPropertyOrDefault(ServiceRemoteInclusion.HEADERS, JsonDocument.emptyDocument());
```
derklaro: Delayed until #1178, this should use `Document.emptyDocument()` after.

derklaro (approved):

PR #1199, opened 2023-04-15
feat: implement a method to delete a directory from a template.

### LINKED ISSUE #1198: Add a method to delete entire directory trees from templates
### Describe the new feature

This new feature should encompass a method called `deleteDirectory` that can be used to delete an entire directory including all its contained files from a given template.

### Why do you need this feature?

Currently it is only possible to delete singular files from a template.
This works fine but may leave empty and unused directories behind.  
In addition to that, deleting entire directory trees is very inefficient when deleting each file separately.

### Alternatives

The only other way of removing the remaining directories would be to manually delete those files.  
Developers of modules cannot do that and would need to instruct an administrator to do this. This would leave developers at the mercy of attentive administrators/users.

### Other

Requested on Discord by the following messages: https://discord.com/channels/325362837184577536/818777626663321671/1096849346735120434

I am currently working on this feature request.

### Issue uniqueness

GiantTreeLP:
### Motivation

See #1198.

### Modification

This change allows developers to directly delete a directory and all its containing files and directories from a given service template.

As an additional functionality, this method is also exposed to the HTTP Rest API much like the file deletion method already is.

### Result

Developers can now delete entire directory trees from a given template.

##### Other context

Closes #1198.

Local template storage has already been tested on my machine, I will check whether SFTP and S3 storage can be tested. If not, I may seek help from your end.

### DIFF
--- a/driver/src/main/java/eu/cloudnetservice/driver/template/TemplateStorage.java
+++ b/driver/src/main/java/eu/cloudnetservice/driver/template/TemplateStorage.java
@@ -221,6 +221,17 @@ public interface TemplateStorage extends AutoCloseable, Named {
   boolean deleteFile(@NonNull ServiceTemplate template, @NonNull String path);
 
   /**
+   * Deletes the given directory at the given path in the given template in this storage.
+   * This method recursively deletes all files and directories inside the given directory.
+   *
+   * @param template the template in which the directory to delete is located in.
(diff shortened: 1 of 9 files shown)


0xIO32 (approved):
Github said "At least 1 approving review is required to merge this pull request.", so i made one

PR #1210, opened 2023-04-23
chore: upgrade WaterDog to v2

derklaro:
### Motivation
WaterdogPE version 2 was released and we should upgrade to this new version as the 1.X will no longer receive updates.

### Modification
Update usages of Waterdog classes from 1.X to 2.X & update configuration file.

### Result
Waterdog v2 can now be used with CloudNet.

#### Other Context
- [ ] Wait for Waterdog to make v2 builds available from their Jenkins


tunikakeks:
I have some issues with the Bridge, it got updated correctly from the waterdog2 branch. Self-built and downloaded from the launchermeta repo

image

PR #1221, opened 2023-05-09
fix: the launchers memory should not effect the node memory

0utplay:
### Motivation
The way our launcher works is that all jvm options of the launcher get passed down to the started node instance. Therefore increasing the memory in one of the start files would result in the memory increasing for two different processes.
1. The launcher process it self start is started by the start file
2. The memory of the new node process started by the launcher

So that means that the memory consumption is doubled.

### Modification
Made sure to remove the -xmx and -xms options of the launcher before passing the remaining options to the node and introduced a new option in the launcher.cnl.
The new option `cloudnet.node.memory` is the new way to set the memory of the node (in megabytes).

### Result
The memory is not occupied twice and the node memory is not dependant on the launcher.


launcher/java17/src/main/java/eu/cloudnetservice/launcher/java17/ApplicationBootstrap.java
```diff
@@ -66,8 +68,18 @@ public static void bootstrap(
     Set<String> arguments = new LinkedHashSet<>();
     arguments.add(resolveJavaExecutable());
 
+    // remove the xms and xmx options as we want to set them later on
+    var filteredInputArguments = RUNTIME_MX_BEAN.getInputArguments().stream().filter(s -> {
```
derklaro: Please rename the lambda argument to something more descriptive

derklaro (changes requested):

derklaro (approved):

PR #1225, opened 2023-05-13
fix: make sure to create parent directories in sftp

0utplay:
### Motivation
When creating a new file on a sftp storage in a non existant directory the file creation will fail instead of creating the parent directories. This will result in an exception when installing a template on a sftp storage due to the creation of the config/sponge/sponge.conf file.

### Modification
Made sure to create the directories before opening an outputstream or creating a file in a sftp template storage. Furthermore the tests were extended to cover the creation of files in non existant directories

### Result
The parent directories are created.


modules/storage-sftp/src/main/java/eu/cloudnetservice/modules/sftp/SFTPTemplateStorage.java
```diff
@@ -309,6 +313,17 @@ public void close() throws IOException {
     }, null);
   }
 
+  protected void createParentDirectory(
```
derklaro: nit: createParentDirectories is more fitting here

derklaro (changes requested):

derklaro (approved):

PR #1262, opened 2023-06-20
fix: draining causes jvm halt instead of a stop

0utplay:
### Motivation
When enabling draining on a node and stopping all already running services the node should be stopped. Currently instead of stopping the node we cause the jvm to halt therefore the node does not stop correctly and does not exit at all.

### Modification
Used the shutdown method in the shutdown handler to stop the node and exit.

### Result
The node exits successfully.

### DIFF
--- a/node/src/main/java/eu/cloudnetservice/node/Node.java
+++ b/node/src/main/java/eu/cloudnetservice/node/Node.java
@@ -461,11 +461,14 @@ public final class Node {
   @Inject
   @Order(10_000)
   private void installShutdownHook(@NonNull Provider<ShutdownHandler> shutdownHandlerProvider) {
-    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+    var shutdownHookThread = new Thread(() -> {
       // get the shutdown handler instance & execute the shutdown process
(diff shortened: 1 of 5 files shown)


node/src/main/java/eu/cloudnetservice/node/Node.java
```diff
@@ -461,11 +461,14 @@ private void scheduleNodeUpdateTasks(
   @Inject
   @Order(10_000)
   private void installShutdownHook(@NonNull Provider<ShutdownHandler> shutdownHandlerProvider) {
-    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+    var shutdownHookThread = new Thread(() -> {
```
derklaro: Thread has a constructor that takes the name. I would suggest using that:
```java
    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
      // get the shutdown handler instance & execute the shutdown process
      var shutdownHandler = shutdownHandlerProvider.get();
      shutdownHandler.shutdown();
    }, ShutdownHandler.SHUTDOWN_THREAD_NAME));
```

derklaro (commented):
small nit

derklaro (approved):

PR #1264, opened 2023-06-27
fix: CORS Header were not present on preflight and added custom headers to the configuration

### LINKED ISSUE #1224: Allowing CORS-Requests
### Describe the new feature

I cannot use the REST-API Module due to CORS. 
Browsers and the applications below (js) do not allow any non-cors request to include a Authorization: Bearer header.

Therefore, I need the "Access-Control-Allow-Origin" Header set to * on all endpoints.

### Why do you need this feature?

I cannot interface with the REST-API properly in browser-environments due to CORS. Therefore, any JS that makes requests to the CloudNet-REST-API Module cannot authenticate.

### Alternatives

If not possible to add due to restrictions (e.g. cluster communication etc.), it would be sufficent to have a Proxy that at least responds with the "Access-Control-Allow-Origin" header set to *, so that the cluster could just continue using the "unsafe" API and all "Public-API calls" could occur on a different port.

### Other

_No response_

### Issue uniqueness

0xIO32:
### Motivation

Just wanted to fix that the CORS Header weren't present on the preflight. I also added the configuration to add custom headers and removed a duplicated registration of the V2HttpHandlerModule.class.

### Modification

Adds the Annotation "ApplyHeaders" with the HeaderAnnotationExtension, which will be used in the CloudNet-Rest module and the CloudNet-Bridge module.
It also makes a few changes the restConfiguration. For example now you can add custom headers.
It also removes this line and the restConfiguration field and constructor parameter, because it isn't needed anymore in V2HttpHandler (and maybe some other classes):
`.header("Access-Control-Allow-Origin", this.restConfiguration.corsPolicy());`

### Result

Now CORS shouldn't be a problem anymore.

##### Other context

Fixes https://github.com/CloudNetService/CloudNet-v3/issues/1224


MarkusTieger:
Notice: requires added the following headers:
Access-Control-Allow-Methods
Access-Control-Allow-Credentials
Access-Control-Allow-Headers

thxrben:
I found an issue with your commit:

PREFLIGHT Requests respond with 404, even if the correct path is given.
Therefore, the Server does not allow the POST/GET request itself and is being turned down.


I merged your and the Cloudnet Repos nightly branches and build & installed the custom jar as dev mode without updates.
The Rest, Bridge and Sign Modules were replaced with the new build too. (https://github.com/Gandalf1783/CloudNet-v3)

Could you take a look at this?

EDIT: The current policies in the config.json are default (e.g. "*", with all headers allowed)
EDIT 2: The headers are added when not having to use cors (In this case, I used the react-proxy setting).

grafik

0xIO32:
I tryed to reproduce, but the headers were present. I even used your request headers from your screenshot.

So you are sure, that "var cloudnet.dev true" is present in the launcher.cnl?

Test layout:
image


Try to reproduce:
image

thxrben:
Yes I am sure. 
The launcher.jar and cnl are both named cloudnet.jar/cnl and the .cnl forbids auto updates and is in dev-mode.

I have replaced the bridge, the launcher aswell as the rest module.
This should be everything needed, right?

I have cloned your repo and build it using ./gradlew -x test (from nightly) and pointed my fetch-requests to my local instance.
Still, no headers in sight.

I also tried the curl request.
My Cloudnet Instance:
image

The Curl request:
`curl -v --request OPTIONS 'http://127.0.1.1:2812/api/v2/auth' -H 'Accept: */*' -H 'Accept-Encoding: gzip,deflate' -H 'Access-Control-Request-Headers' -H 'Access-Control-Request-Method: POST' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Host: http://127.0.1.1:2812' -H 'Origin: http://google.com' -H 'Referer: http://google.com/' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/114.0'`

image

thxrben:
I have to excuse myself:
Due to the dev-mode not working correctly, CloudNet updated the modules itself and replaced the compiled jar by myself with another older one.

Therefore the requests never used your Preflight fix and therefore were never correct.

I cant test it, I am currently taking a look at dev mode again and will test this ASAP.

EDIT: Spelling

MarkusTieger:
And why is there a "Loading checksums", "Skipping download" entries in the logs? There shouldn't be present.
And the latest version of my fork/pr is "4.0.0-RC10-SNAPSHOT-5763669a" and not "4.0.0-RC9-f6ca4c38".
I think you are running this commit and not mine from the pr: https://github.com/CloudNetService/CloudNet-v3/commit/f6ca4c38b18ebff6a8a58fbcdd34b87dbadc39fb

image

thxrben:
Like i said, I had issues with the dev mode.

I misplaced the cloudnet.jar and launcher.jar in my install and therefore it auto updated everything while it shouldn't.

Your CORS fix works fine btw, tested it successfully now :)

docimin:
Hey.

Cors issues should be fixed within one of the next updates most likely, as there will be a REST API rework.

0utplay:
As mentioned, we have done some improvements to the REST api. I will close this, as the improvements will resolve this

PR #1330, opened 2023-11-03
fix: tablist order

ItzN00bPvP:
### Motivation
The Tablist wasn't in the expected order. And i wanted to know why.

### Modification
use the sortID to prefix the teamName and not the length of the sortid

### Result
The Tablist is in the expected order according to the sortId


derklaro:
no longer needed: permission module, simplenametags & chat plugin were removed. thanks for your effort nevertheless

PR #1353, opened 2024-02-21
feat: extend websocket support

matserix:
### Motivation
JavaScript WebSocket API does not support adding authorization headers making the live console endpoint unusable. The only controllable header is `Sec-WebSocket-Protocol`, which is not convenient to use and support.

### Modification/Result
Methods or parameter annotated with [TokenAuth](https://github.com/NitrinCloud/CloudNet-v3/blob/http-websocket/node/src/main/java/eu/cloudnetservice/node/http/annotation/TokenAuth.java) allows to either add the generated token as a bearer token or as a query parameter. 

I made the decision to create a third annotation because this will still force a user to use the correct header if the endpoint won't upgrade to a websocket connection later on.


0utplay:
Hi, sorry for the late reply on this. We are currently rewriting the rest module and have solved this issue there. Thanks for you effort.

PR #1399, opened 2024-05-16
fix: properly extract kick messages from downstream services

0utplay:
### Motivation
When players are kicked from e.g. a Lobby service, the message the Lobby kicked them with is lost. This is a problem when having issues with version mismatches or plugins kicking someone from a fallback as the kick message is lost there too.

### Modification
When getting kicked from a service and no other fallback is available the player will be kicked with the kick message of the downstream service. If there is no kick message, the newly introduced "server-kick-no-other-hub" message will be used.

Also if a player joins and there is no fallback and the "proxy-join-disconnect-because-no-hub" is not set to `null`, the player will be disconnected with the set message. If the message is not set, cloudnet won't do anything. This ensures that other plugins can overwrite the decision.

### Result
Kick reasons are properly displayed.


Dangles:
I ask the developers to fix this problem as soon as possible and add this Pull because plugins on punishment, like LibertyBans and LiteBans outside the proxy system, don't work because of it. LibertyBans 100% does not work. Other plugins on my servers that kick players with a reason don't work either, there are conflicts with messages from CloudNet, for example “proxy-join-disconnect-because-no-hub” in Bridge. If some plugin outside the proxy tries to ban/kick or send a message to the client, the message “proxy-join-disconnect-because-no-hub” will appear or in some cases account lockout may not happen. 

I don't understand how to make my server support bans/kicks without spending money on my Cloudnet plugin to fix such a serious incompatibility issue with other plugins. That said, I don't want to give up Cloudnet for this. Hopefully the problem will be resolved.

PR #1410, opened 2024-05-22
feat: add sponge 1.20.6 downloads

RobotHanzo:
### Motivation
Sponge now supports 1.20.6

### Modification
Add 1.20.6 for sponge

### Result
Allows PR #1395 to be complete and mergable

##### Other context

Wraps up PR #1395

### DIFF
--- a/node/src/main/resources/files/versions.json
+++ b/node/src/main/resources/files/versions.json
@@ -480,11 +480,11 @@
             },
             "fetchOverSpongeApi": true,
             "artifact": "spongevanilla",
-            "mcVersion": "1.20.4"
+            "mcVersion": "1.20.6"
           }
         },
         {
-          "name": "1.20.4",
+          "name": "1.20.6",
           "properties": {
             "copy": {
               "libraries/.*": "/",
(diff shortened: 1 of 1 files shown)


0utplay:
Hi, have you tested sponge 1.20.6 with all modules that have sponge support?

RobotHanzo:
> Hi, have you tested sponge 1.20.6 with all modules that have sponge support?

I haven't as sponge's API kept giving time outs and 524 for all versions, it'd be great if someone else can test it
`Caused by: java.io.IOException: Server returned HTTP response code: 524 for URL: https://repo.spongepowered.org/service/rest/v1/search/assets?md5=e3899bcf5a638
22d9c41dff001559921aven.groupId=org.spongepoweredaven.artifactId=configurate-yamlaven.baseVersion=4.1.2aven.extension=jar
`

0utplay:
Okay, so we'll have some problems with our platform inject because of the new sponge api version. I think I will need some time because at this point we need to remove cloudperms in order to proceed with sponge

PR #1413, opened 2024-05-25
feat: toggle for remote inclusion caches

0utplay:
### Motivation
Some users use remote inclusions to always download latest files of certain plugins. The automatic caches makes that a bit hard.

### Modification
Added a new `cacheFiles` option for remote inclusions and we only cache files when the property is enabled.

### Result
The caching can be toggled using the `cacheFiles` option


node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
```diff
@@ -394,6 +394,11 @@ public void includeWaitingServiceInclusions() {
         FileUtil.copy(destination, target);
         // we've installed the inclusion successfully
         this.installedInclusions.add(inclusion);
+
+        // we don't want to keep the temporary file if caching is disabled
+        if (!inclusion.cacheFiles()) {
+          FileUtil.delete(destination);
+        }
```
derklaro: I'm wondering: wouldn't it be easier to just download the file into the service directory rather than downloading to a temp dir, copying the file & deleting it?

copyandexecute:
sieeeht gut aus, wird uns freuen

copyandexecute:
passt oder

node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
```diff
@@ -766,6 +771,11 @@ protected void prepareService() {
       configuration.privateKeyPath() == null ? null : wrapperDir.resolve("privateKey"));
   }
 
+  protected void downloadInclusionFiles(@NonNull GetRequest request, @NonNull Path destination) {
```
derklaro: Is it indeted that the method name is plural but really only downloads one file?

derklaro (dismissed):
Look good in general, just one thought: would it be better to implement equals/hashCode for inclusions specifically? I'm wondering if cacheFiles anwhere contributes anything, that would make it a requirement to include in these methods...

driver/src/main/java/eu/cloudnetservice/driver/service/ServiceRemoteInclusion.java
```diff
@@ -22,18 +22,21 @@
...
  * An inclusion which can be added to a service and will download a file from the specified url to the given
  * destination. This can be anything for example a plugin or configuration file.
  *
  * @since 4.0
  */
-@EqualsAndHashCode
 public final class ServiceRemoteInclusion implements DefaultedDocPropertyHolder, Cloneable {
 
+  public static final String NO_CACHE_STRATEGY = "NO_CACHE";
+  public static final String KEEP_UNTIL_RESTART_STRATEGY = "KEEP_UNTIL_NODE_RESTART";
```
derklaro: should they be upper case? feels like lower case would be better?
and maybe renaming them to `none` & `until-node-restart` (dropping the cache_ prefix/suffix)?

driver/src/main/java/eu/cloudnetservice/driver/service/ServiceRemoteInclusion.java
```diff
@@ -134,6 +157,22 @@ private ServiceRemoteInclusion(@NonNull String url, @NonNull String destination,
     }
   }
 
+  @Override
```
derklaro, line 162: inheritDoc missing

driver/src/main/java/eu/cloudnetservice/driver/service/ServiceRemoteInclusion.java
```diff
@@ -143,6 +182,7 @@ public static class Builder {
 
     protected String url;
     protected String destination;
+    protected String cacheStrategy = "NO_CACHE";
```
derklaro: this should probably use the constant instead

driver/src/main/java/eu/cloudnetservice/driver/service/ServiceRemoteInclusion.java
```diff
@@ -172,6 +212,18 @@ public static class Builder {
...
     }
 
+    /**
+     * Sets whether the downloaded file will be cached on the node. If set to {@code true} the file is downloaded once
+     * and cached until the cloud is restarted.
+     *
+     * @param cacheStrategy whether to cache the downloaded file.
+     * @return the same instance as used to call the method, for chaining.
+     */
+    public @NonNull Builder cacheStrategy(@NonNull String cacheStrategy) {
```
derklaro, line 227: Documentation no longer reflects what the method is actually doing. I'm not sure if this should mention the service registry right away or if this should be changed in the followup accordingly

derklaro (changes requested):

derklaro (approved):

PR #1422, opened 2024-06-25
chore: Increase timeout for process termination

GiantTreeLP:
### Motivation
Current Java Edition Minecraft servers, especially those that are forks of the vanilla Minecraft server software, take some time to properly stop and save all resources, i.e. worlds, inventories and other state to disk. Prematurely terminating those processes leads to worlds not being saved at all, world corruption due to the interrupted save process and unloadable worlds. In addition this leads to lock files being left that don't belong to any process anymore.

### Modification
Increase the timeout to 60 seconds. This gives servers more time to properly store all files and shutdown whilst still cleaning up servers that may have become stuck whilst shutting down.

### Result
Server processes are now given more time to properly shutdown.

##### Other context
See the discussion in Discord starting from [here](https://discord.com/channels/325362837184577536/818777626663321671/1255228564832260127).

### DIFF
--- a/node/src/main/java/eu/cloudnetservice/node/config/JsonConfiguration.java
+++ b/node/src/main/java/eu/cloudnetservice/node/config/JsonConfiguration.java
@@ -212,7 +212,7 @@ public final class JsonConfiguration implements Configuration {
     if (this.processTerminationTimeoutSeconds <= 0) {
       this.processTerminationTimeoutSeconds = ConfigurationUtil.get(
         "cloudnet.config.processTerminationTimeoutSeconds",
-        5,
+        60,
         Integer::parseInt);
     }
 


DasBabyPixel:
I feel like this should be configurable instead of fixed 60s (or 5s)
Best would be a setting for the task imo. Some lobby template servers only need 5s and should be killed after, some heavily modded static server might need more. Yet you also don't want to wait for a whole minute for broken lobby server to be killed.

derklaro:
Yes, the configuration should be available per-service, but that's not a simple task to do as the task configuration is not available to the service, unless we provide the timeout as a variable in the service configuration or retrieve the task from the task name provided to the service configuration.

DasBabyPixel:
> Yes, the configuration should be available per-service, but that's not a simple task to do as the task configuration is not available to the service, unless we provide the timeout as a variable in the service configuration or retrieve the task from the task name provided to the service configuration.

what's the problem with providing it as a variable in the service configuration?

derklaro:
The thing is: sure, we could just throw the variable in there and move on, but there are more places that have variables that cannot be set per service, and need to be accessed via the task (which shouldn't be the case). One example is the required permission for a task or the smart configuration. So imo it's better to delay this a bit and find a better solution on how we can handle the variables and then merge this :)

PR #1429, opened 2024-07-01
WorldEdit transformers for 1.8.8 + give more control to transformers (wrapper)

DasBabyPixel:
### Motivation

A few servers running bukkit 1.8.8 still use worldedit. Only since the change to modern java on those old versions, worldedit stopped to work. This PR fixes worldedit and FAWE on 1.8.8.
The necessary class transformers also uncovered a problem in the class transformer api. Converting a `ClassNode` back to a `byte[]` was not possible (in some cases), because of `COMPUTE_FRAMES|COMPUTE_MAXS`.

Some might also want to make use of the wrapper transformer API in wrapper modules, so the custom `byte[]` creation might come in handy.
As an example, I have another project that needs transforming of classes, but uses a `byte[]` instead of asm.
The new API allows something like this:
```
@Override
public byte[] toByteArray(ClassNode classNode) {
    var bytes = Transformer.super.toByteArray(classNode);
    return CustomTransformer.transform(bytes);
}
```

### Modification

Added transformers for WorldEdit / FAWE
Added a custom opt-in serializer option for transformers.
Added better logging for transformer failures.

### Result

FAWE / WorldEdit supported.
Proper handling of transformer errors / soft failure - not doing this causes the JVM to freeze on errors iirc.
Give more control to transformers, without changing their default behaviour.

##### Other context

This is something I concocted some ago and kept in my fork, thought this might be of interest.
I did also test worldedit/fawe on 1.20 to make sure this doesn't cause any problems. Didn't notice anything, but as always with transformers, future changes on worldedit (they'd have to do something odd, but still) could cause the transformer to break worldedit functionality.

If the worldedit transformers aren't merged, I'd still love to see the API to give more control to transformers merged, because I use the API changes privately. WorldEdit transformers could be re-added by a wrapper module, but only with the more powerful api.

### DIFF
--- a/wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEConfigTransformer.java
+++ b/wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEConfigTransformer.java
@@ -27,6 +27,13 @@ public class FAWEConfigTransformer implements ClassTransformer {
   private static final String MN_SET_ACCESSIBLE = "setAccessible";
 
   /**
+   * Constructs a new instance of this transformer, usually done via SPI.
+   */
+  public FAWEConfigTransformer() {
+    // used by SPI
+  }
+
+  /**
    * {@inheritDoc}
    */
   @Override
--- a/wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEConfigTransformer.java
+++ b/wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEConfigTransformer.java
@@ -20,8 +20,15 @@ import eu.cloudnetservice.wrapper.transform.ClassTransformer;
 import java.lang.classfile.ClassTransform;
 import java.lang.classfile.CodeTransform;
 import lombok.NonNull;
+import org.jetbrains.annotations.ApiStatus;
 
-public class FAWEConfigTransformer implements ClassTransformer {
+/**
+ * A transformer implementation that disables the {@code setAccessible} method in the FAWE Config on old FAWE versions.
+ * This is due to the fact that the method uses illegal reflection in the attempt to set a final field which is not
+ * possible anymore on newer java versions.
+ */
+@ApiStatus.Internal
+public final class FAWEConfigTransformer implements ClassTransformer {
 
   private static final String CNI_CONFIG = "com/boydti/fawe/config/Config";
   private static final String MN_SET_ACCESSIBLE = "setAccessible";
--- a/wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEConfigTransformer.java
+++ b/wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEConfigTransformer.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2019-2024 CloudNetService team & contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package eu.cloudnetservice.wrapper.transform.bukkit;
+
+import eu.cloudnetservice.wrapper.transform.ClassTransformer;
+import java.lang.classfile.ClassTransform;
+import java.lang.classfile.CodeTransform;
+import lombok.NonNull;
+
+public class FAWEConfigTransformer implements ClassTransformer {
+
+  private static final String CNI_CONFIG = "com/boydti/fawe/config/Config";
+  private static final String MN_SET_ACCESSIBLE = "setAccessible";
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public @NonNull ClassTransform provideClassTransform() {
+    CodeTransform codeTransform = (builder, _) -> builder.return_();
+    return ClassTransform.transformingMethodBodies(
+      mm -> mm.methodName().equalsString(MN_SET_ACCESSIBLE),
+      codeTransform);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public @NonNull TransformWillingness classTransformWillingness(@NonNull String internalClassName) {
+    var isFaweConfig = CNI_CONFIG.equals(internalClassName);
+    return isFaweConfig ? TransformWillingness.ACCEPT_ONCE : TransformWillingness.REJECT;
+  }
+}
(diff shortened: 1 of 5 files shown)


0utplay:
I like the idea, but this might have to wait until we've replaced the networking and asm

0utplay:
If you are still willing to do these changes: You have to use the new classfile api

DasBabyPixel:
Can I give the ClassTransform class the option (a default method, or sth else) of manually modifying the byte[] of the class after using java's api? I mostly did the whole thing to be able to manually apply papermc transforms to classes contained in a wrapper module (thus not transformed by paper's classloader).
To do something like this: 
```
@Override
public byte[] toByteArray(ClassNode classNode) {
    var bytes = Transformer.super.toByteArray(classNode);
    return CustomBytecodeModifier.modify(bytes);
}
```

DasBabyPixel:
Maybe I'm also missing a way to do the same thing using java's new api. If so, let me know

0utplay:
cc @derklaro

derklaro:
The wrapper transformer api is build to transform *classes* not to implement custom transformers which, imo, is done with the changes made to the api recently.

If you want to transform the original class bytes of a class, I suggest to either write your own java agent and attach it to register custom transformers, or, another way would be to use a (kinda) hacky approach to get the original class bytes, pass them to your transformer like this:
```java
static class TestClassTransform implements ClassTransform {

  @Override
  public void atStart(ClassBuilder builder) {
    var classModel = builder.original().orElseThrow(); // transforming, so this is always present
    var classReader = (ClassReader) classModel.constantPool();
    var originalClassBytes = classReader.readBytes(0, classReader.classfileLength());
    // pass to custom transformer, parse the class file & extract the instructions to call from it
  }

  @Override
  public void accept(ClassBuilder builder, ClassElement element) {
    // builder.accept(element); - use this to accept the old class
    // or use the extracted elements above and pass them to the builder
  }
}
```

DasBabyPixel:
> The wrapper transformer api is build to transform _classes_ not to implement custom transformers which, imo, is done with the changes made to the api recently.
> 
> If you want to transform the original class bytes of a class, I suggest to either write your own java agent and attach it to register custom transformers, or, another way would be to use a (kinda) hacky approach to get the original class bytes, pass them to your transformer like this:
> 
> ```java
> static class TestClassTransform implements ClassTransform {
> 
>   @Override
>   public void atStart(ClassBuilder builder) {
>     var classModel = builder.original().orElseThrow(); // transforming, so this is always present
>     var classReader = (ClassReader) classModel.constantPool();
>     var originalClassBytes = classReader.readBytes(0, classReader.classfileLength());
>     // pass to custom transformer, parse the class file & extract the instructions to call from it
>   }
> 
>   @Override
>   public void accept(ClassBuilder builder, ClassElement element) {
>     // builder.accept(element); - use this to accept the old class
>     // or use the extracted elements above and pass them to the builder
>   }
> }
> ```

Allright, I'm gonna finish the transformers in this PR (they don't need byte[] manipulation). Going to try your suggestion, if an issue arises, I'm gonna open a new issue or PR

derklaro:
I'm happy to pull in the WorldEdit transformers. If there are issues feel free to contact me on Discord :)

DasBabyPixel:
Finished the transformers, tested `//pos1 //pos2 //set <block>` on paper 1.8.8, 1.12.2, 1.16.5 and 1.21. 
See commit message for exact versions used

wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEConfigTransformer.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package eu.cloudnetservice.wrapper.transform.bukkit;
+
+import eu.cloudnetservice.wrapper.transform.ClassTransformer;
+import java.lang.classfile.ClassTransform;
+import java.lang.classfile.CodeTransform;
+import lombok.NonNull;
+
+public class FAWEConfigTransformer implements ClassTransformer {
```
derklaro:
```suggestion
@ApiStatus.Internal
public final class FAWEConfigTransformer implements ClassTransformer {
```
And I would really appreciate if you could add a small javadoc comment to all classes describing what the transformer is doing.

wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEReflectionUtilsTransformer.java
```diff
@@ -0,0 +1,72 @@
...
+    ConstantDescs.CD_Object);
+
+  /**
+   * Constructs a new instance of this transformer, usually done via SPI.
+   */
+  public FAWEReflectionUtilsTransformer() {
+    // used by SPI
+  }
+
+  @Override
```
derklaro, line 57: `@inheritDoc` is missing on these

wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEReflectionUtilsTransformer.java
```diff
@@ -0,0 +1,72 @@
...
+  public @NonNull ClassTransform provideClassTransform() {
+    CodeTransform codeTransform = (builder, e) -> {
+
+      // field.setAccessible(true);
```
derklaro:
```suggestion
    CodeTransform codeTransform = (builder, _) -> {
      // field.setAccessible(true);
```

wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/bukkit/FAWEWorldEditDownloadURLTransformer.java
```diff
@@ -0,0 +1,63 @@
...
+      if (element instanceof ConstantInstruction.LoadConstantInstruction loadConstant) {
+        if (loadConstant.opcode() == Opcode.LDC) {
+          if (loadConstant.constantValue().equals(OLD_DOWNLOAD_URL)) {
+            builder.ldc(NEW_DOWNLOAD_URL);
+            return;
+          }
+        }
+      }
+      builder.with(element);
+    };
```
derklaro: we had a recent incident where the constant was loaded using `LDC_W`, i would personally just check if the constant value is a string instead. Also, merging all these ifs should be better:
```suggestion
    CodeTransform codeTransform = (builder, element) -> {
      if (element instanceof ConstantInstruction.LoadConstantInstruction inst
        && inst.constantValue() instanceof String string
        && string.equals(OLD_DOWNLOAD_URL)) {
        builder.ldc(NEW_DOWNLOAD_URL);
        return;
      }
      builder.with(element);
    };
```

derklaro (changes requested):
A few minor nits, but generally looking good

DasBabyPixel:
Done

derklaro (approved):
didn't test this, but code LGTM

0utplay:
I will do some small testing later and merge afterwards. Thank you!

0utplay:
So I did face an issue using fawe and the //set Command. 
image

But worldedit was working fine and fawe is also working and I think you can argue that people would have to live with that

DasBabyPixel:
> So I did face an issue using fawe and the //set Command. image
> 
> But worldedit was working fine and fawe is also working and I think you can argue that people would have to live with that

I didn't notice that because I had anti xray turned off. Gonna see if there is an easy fix, otherwise oh well...
I do want to point out that this also happens with java 8 on a pure paper server without anything else

DasBabyPixel:
After further investigation, this happens because timings are stopped while async (duh). 
FAWE sends chunk updates async and the PacketPlayOutMapChunk hides blocks async, causing this. 
This is not actually an error but rather a warning that also up screws timings a little

image

I see 3 options
Option 1: Disable timings in config
Option 2: Disable anti xray in config
Option 3: Add transformer disabling either

I think it is best to ignore this and tell ppl to change their config to disable timings. We have spark for a reason.

PR #1432, opened 2024-07-02
chore: use classdesc instead of asm type

0utplay:
### Motivation
 We want to get rid of asm

### Modification
 Removed the usage of asms type and replaced it with the java alternative.

### Result
 Asm is not used anymore in this context


derklaro (approved):

DasBabyPixel:
Is asm also going to be removed in transformers?

derklaro:
> Is asm also going to be removed in transformers?

Yes, that is the plan.

PR #1444, opened 2024-07-12
chore: replace asm in class transformers with jdk22 classfile api

derklaro:
### Motivation
We're using JDK22 which includes a new class file api which we can use to replace existing usages of ASM. While most usages are already replaced, the transformers that the wrapper is using to make some old minecraft versions compatible with new jdk versions are still using ASM. This causes trouble with all kinds of platforms that also include ASM on their own (such as Spigot/Paper, Fabric, Forge, ...).

### Modification
Drop the ASM dependency completely from the project by replacing the ASM usages in the wrapper transformers with the jdk class file api (the last usage of ASM).

### Result
The wrapper transformers are using the new jdk class file api which will prevent issues with conflicting ASM versions going forward. This also drops the dependency on ASM.


wrapper-jvm/src/main/java/eu/cloudnetservice/wrapper/transform/ClassTransformer.java
```diff
@@ -0,0 +1,74 @@
...
+   * @since 4.0
+   */
+  enum TransformWillingness {
+
+    /**
+     * Indicates that the class transformer wants to transform the given target class.
+     */
+    ACCEPT,
+    /**
+     * Indicates that the class transformer wants to transform the given target class and has doesn't want to transform
```
0utplay: The sentence somehow does not make sense

0utplay (approved):

PR #1450, opened 2024-07-13
chore: remove Task and replace it with CompletableFuture

0utplay:
### Motivation
When mixing different apis and using our Task implementation the problem occurs that after using any of the thenApply features the Task is a CompletableFuture and no Task anymore. 

### Modification
Removed the Task and only left some util methods for working with CompletableFuture

### Result
No issues when using CompletableFuture


driver/src/main/java/eu/cloudnetservice/driver/template/defaults/RemoteTemplateStorage.java
```diff
@@ -96,28 +97,42 @@ public boolean deployDirectory(
    */
   @Override
   public boolean deploy(@NonNull ServiceTemplate target, @NonNull InputStream inputStream) {
+    return this.deployAsync(target, inputStream).join();
```
derklaro: should probably use TaskUtil.getOrDefault

driver/src/main/java/eu/cloudnetservice/driver/template/defaults/RemoteTemplateStorage.java
```diff
@@ -96,28 +97,42 @@ public boolean deployDirectory(
...
-      .join();
+      .thenApply(status -> status == TransferStatus.SUCCESS);
   }
 
   /**
    * {@inheritDoc}
    */
   @Override
   public @Nullable InputStream zipTemplate(@NonNull ServiceTemplate template) {
+    return this.zipTemplateAsync(template).join();
```
derklaro: should probably use TaskUtil.getOrDefault

driver/src/main/java/eu/cloudnetservice/driver/template/defaults/RemoteTemplateStorage.java
```diff
@@ -184,11 +199,18 @@ public boolean deploy(@NonNull ServiceTemplate target, @NonNull InputStream inpu
     @NonNull ServiceTemplate template,
     @NonNull String path
   ) throws IOException {
+    return this.newInputStreamAsync(template, path).join();
```
derklaro: should probably use TaskUtil.getOrDefault

derklaro (changes requested):

derklaro (approved):

PR #1453, opened 2024-07-13
feat: set autoDeleteOnStop using the tasks command

RobotHanzo:
### Motivation
`tasks task <name> set` was missing the ability to edit `autoDeleteOnStop`

### Modification
Added the command `tasks task <name> set autoDeleteOnStop <enabled>`

### Result
`tasks task <name> set` can now modify `autoDeleteOnStop`, without having to edit the json file by hand

##### Other context
N/A


0utplay:
I'm not really sure. When I implemented the commands for v4 I left that one out on purpose because a lot of users do net get what this means and confuse it with `staticServices`. At least until there is no proper documentation I would leave it like it is

RobotHanzo:
> I'm not really sure. When I implemented the commands for v4 I left that one out on purpose because a lot of users do net get what this means and confuse it with `staticServices`. At least until there is no proper documentation I would leave it like it is

I thought leaving it out from tasks setup would be enough (as this is probably what most starters start with)
Advanced users probably need a convenient method to configure this though (in my opinion)

RobotHanzo:
> I'm not really sure. When I implemented the commands for v4 I left that one out on purpose because a lot of users do net get what this means and confuse it with `staticServices`. At least until there is no proper documentation I would leave it like it is

I thought leaving it out from tasks setup would be enough (as this is probably what most starters start with)
Advanced users probably need a convenient method to configure this though (in my opinion)

node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
```diff
@@ -385,6 +385,21 @@ public void renameTask(
...
+  @CommandMethod("tasks task <name> set autoDeleteOnStop <enabled>")
+  public void setAutoDeleteOnStop(
+    @NonNull CommandSource source,
+    @NonNull @Argument("name") Collection<ServiceTask> tasks,
+    @Argument("enabled") @Liberal boolean enabled
+  ) {
+    this.applyChange(
+      source,
+      tasks,
+      ServiceTask.Builder::maintenance,
```
0utplay: I do not think this is right here

0utplay (changes requested):

node/src/main/java/eu/cloudnetservice/node/command/sub/TasksCommand.java
RobotHanzo: > I do not think this is right here

Forgot to change it, fixed

0utplay (approved):

PR #1454, opened 2024-07-13
feat: A shutting down state in ShutdownHandler

RobotHanzo:
### Motivation
Before this PR, there was no way of figuring if cloudnet is shutting down or not outside of internal cloudnet packages (without using reflections), causing custom load balancing modules unable to determine whether it should start new services.

### Modification
This PR exposes a proper function to determine whether a shutdown is in progress or not.

### Result
Custom modules can now determine whether cloudnet is shutting down or not

##### Other context
N/A


0utplay:
I like the idea but I think it would be best to add a getter to the TickLoop which provides information about the RUNNING state

RobotHanzo:
> I like the idea but I think it would be best to add a getter to the TickLoop which provides information about the RUNNING state

Indeed, I've changed the code

derklaro (approved):
LGTM, thank you

PR #1465, opened 2024-07-26
feat: run bridge post-initialization method at the first server tick

GiantTreeLP:
### Motivation

Certain features of CloudNet react to services updating and thus changing their state to "Running". In the current implementation, this update happens at the time the bridge module/plugin/extension is enabled. This timing is not ideal as a lot of time-consuming tasks, like generating or loading the worlds/dimensions and other modules/plugins/extensions may happen after that. This causes the other components, like the signs module or the NPCs module, to report those services as being already online and ready to join by players, whilst the opposite is actually the case.

### Modification

All currently supported server software platform integrations are augmented, so that `PlatformBridgeManagement#postInit` is called in the first game tick. This tick is only called, once the server is actually initialized, loaded and running. This also means that the server is ready to accept players.

### Result

Amongst other things, signs now show the server for joining only, once the server is actually ready to accept players.

#### Testing

This change has been successfully tested with servers running Paper, Fabric and Limbo.

### DIFF
--- a/modules/bridge/impl/src/main/java/eu/cloudnetservice/modules/bridge/impl/platform/bukkit/BukkitBridgePlugin.java
+++ b/modules/bridge/impl/src/main/java/eu/cloudnetservice/modules/bridge/impl/platform/bukkit/BukkitBridgePlugin.java
@@ -64,7 +64,8 @@ public final class BukkitBridgePlugin implements PlatformEntrypoint {
   public void onLoad() {
     // init the bridge management
     this.bridgeManagement.registerServices(this.serviceRegistry);
-    this.bridgeManagement.postInit();
+
+    this.plugin.getServer().getScheduler().runTask(this.plugin, this.bridgeManagement::postInit);
     // register the bukkit listener
     this.pluginManager.registerEvents(this.playerListener, this.plugin);
   }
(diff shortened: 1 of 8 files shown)


GiantTreeLP:
The forced push fixes an issue where the method reference used as a `Runnable` is not removed from the list of tickable tasks in Fabric.

0utplay:
It is guaranteed that no player can join before the first server tick happens right? Because otherwise this might cause issues with fallbacks and some of our login checks

GiantTreeLP:
For vanilla, ie. Bukkit, Spigot and Paper, I could not absolutely determine that, due to the obfuscated nature of the vanilla Java Edition server.

As I have understood from MinecraftServer.java and DedicatedServer.java, ticking scheduled tasks happens before ticking network connections. I don't know, however, if the login sequence happens on a tick or rather in parallel; for that I did not have access to enough decompiled source code.

I will check with other server software.

DasBabyPixel:
Or set a flag that denies all logins before the logic happens, which in turn also removes the flag. Would work on all platforms, no issues

EDIT: LuckPerms does something very similar

0utplay (approved):

PR #1466, opened 2024-07-29
feat: Configurable log levels

RobotHanzo:
### Motivation
The current log file level have been hard coded to DEBUG, and this causes a huge build up of logs (more the few hundreds of megabytes on my server). Things like docker http requests, minecraft packets are all logged in here, and wouldn't be needed by the average user under normal conditions.

### Modification
Offers a way to modify the log level via system properties, and made INFO the default level of log file level

### Result
No more useless logs

##### Other context
N/A


0utplay:
I have to think about this one. I like having more logging in files because when searching for issues you want to have the information together and not enable debug mode first and reproduce again.

But adding a sys prop is fine.

PR #1478, opened 2024-08-12
fix: ensure to only remove npcs that are currently spawned

0utplay:
### Motivation
When stopping a service and npcs are already removed before the actual disable of the npc plugin the call to remove all npcs might result in a NPE.

### Modification
Ensured to only remove spawned npcs.

### Result
No NPE anymore.


derklaro:
Shouldn't remove be guarded against throwing an exception in case the npc is already removed instead?

derklaro (approved):

PR #1485, opened 2024-08-18
feat: add npc info command

0utplay:
### Motivation
NPCs do not really have an option to display all set options.

### Modification
Added a new command (/cn info) which prints all information about a npc.

### Result
NPC information are available (especially for support)


modules/npcs/src/main/java/eu/cloudnetservice/modules/npc/platform/bukkit/command/NPCCommand.java
```diff
@@ -196,6 +197,57 @@ public boolean onCommand(CommandSender sender, Command command, String label, St
           return true;
         }
 
+        case "info" -> {
+          var npc = this.getNearestNPC(player.getLocation());
+          if (npc == null) {
+            sender.sendMessage("§cNo npc in range found! Make sure the npc you want to edit is in a 5 block radius.");
```
derklaro: Message still says "edit"

modules/npcs/src/main/java/eu/cloudnetservice/modules/npc/platform/bukkit/command/NPCCommand.java
```diff
@@ -614,6 +666,7 @@ public boolean onCommand(CommandSender sender, Command command, String label, St
       return Arrays.asList(
         "create",
         "edit",
+        "info",
```
derklaro: Info is now tab-completable but it's not listed in the help output. Is this command more of a debug thing that shouldn't be listed? In that case I would remove it from tab completion as well.

derklaro (approved):

PR #1490, opened 2024-08-26
refactor: improve service process output reading

derklaro:
### Motivation
Currently service process output streams are read once per second on a single shared thread. This causes issues with delayed logs, especially when the JVM crashes at startup as 99% of the times the JVM crash reason wasn't read from the process error stream. This leaves users with nearly undiagnosable startup issues.

Example log output with an invalid JVM flag before the change:
```
4.0.0-RC11-SNAPSHOT-f66323ca => create by Proxy 1 --start
[26.08 09:26:50.417] INFO : Starting to create 1 services for Proxy
[26.08 09:26:50.641] INFO : CloudService [uniqueId=04789662-d056-48ba-975a-f69993ed56dd task=Proxy name=Proxy-1] has been started
[26.08 09:26:50.642] INFO : The services were created based on the task. They can be managed with the service command
[26.08 09:26:50.703] INFO : CloudService [uniqueId=04789662-d056-48ba-975a-f69993ed56dd task=Proxy name=Proxy-1] has been stopped
```

### Modification
Now the log cache uses virtual threads to read the process output streams. This is very efficient as the whole process stream reading is based on I/O operations which means that the virtual thread can be unmounted until some data is available to be read. This means that even with starting 2 virtual threads per service, there shouldn't be any noticeable performance impacts on the node at all.

Additionally, empty log lines are now ignored in the parent log cache class and there is no need anymore for log caches to implement the handling themselfes. Also a small issue with caching was fixed that caused the cache to still contain 1 item even if the log cache entry limit was set to 0.

### Result
Log output of services is not read immediately once data is available from the input or output stream, making the output more responsive (as it is nearly immediately printed) and also catches everything that is being logged.

Example log output with an invalid JVM flag & this change applied:
```
4.0.0-RC11-SNAPSHOT-b99d7fbc => create by Proxy 1 --start
[26.08 09:21:59.245] INFO : Starting to create 1 services for Proxy
[26.08 09:21:59.289] INFO : CloudService [uniqueId=3a718134-d248-41d4-9f40-75751e1c1d64 task=Proxy name=Proxy-1] has been started
[26.08 09:21:59.289] INFO : The services were created based on the task. They can be managed with the service command
[26.08 09:21:59.295] WARN : [Proxy-1/WARN]: Unrecognized option: -XIReallyKnowWhatIAmDoingISwear
[26.08 09:21:59.296] WARN : [Proxy-1/WARN]: Error: Could not create the Java Virtual Machine.
[26.08 09:21:59.296] WARN : [Proxy-1/WARN]: Error: A fatal exception has occurred. Program will exit.
[26.08 09:22:00.303] INFO : CloudService [uniqueId=3a718134-d248-41d4-9f40-75751e1c1d64 task=Proxy name=Proxy-1] has been stopped
```

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -20,7 +20,7 @@
   "https://checkstyle.org/dtds/configuration_1_3.dtd">
 <module name="Checker">
   <module name="BeforeExecutionExclusionFileFilter">
-    <property name="fileNamePattern" value="(module\-info\.java$)|(.*[\\|\/]fabric[\\|\/].*$)"/>
+    <property name="fileNamePattern" value="(module\-info\.java$)|(.*[\\|\/]fabric[\\|\/].*|(JVM|Dockerized)Service\.java$)"/>
   </module>
   <module name="SuppressionFilter">
     <property default="checkstyle-suppressions.xml" name="file"
--- a/node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
+++ b/node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
@@ -29,6 +29,7 @@ import eu.cloudnetservice.driver.channel.ChannelMessageSender;
 import eu.cloudnetservice.driver.channel.ChannelMessageTarget;
 import eu.cloudnetservice.driver.document.Document;
 import eu.cloudnetservice.driver.event.EventManager;
+import eu.cloudnetservice.driver.event.events.service.CloudServiceLogEntryEvent;
 import eu.cloudnetservice.driver.network.HostAndPort;
 import eu.cloudnetservice.driver.network.NetworkChannel;
 import eu.cloudnetservice.driver.network.buffer.DataBuf;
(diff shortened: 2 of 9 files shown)


checkstyle.xml
```diff
@@ -20,7 +20,7 @@
   "https://checkstyle.org/dtds/configuration_1_3.dtd">
 <module name="Checker">
   <module name="BeforeExecutionExclusionFileFilter">
-    <property name="fileNamePattern" value="(module\-info\.java$)|(.*[\\|\/]fabric[\\|\/].*$)"/>
+    <property name="fileNamePattern" value="(module\-info\.java$)|(.*[\\|\/]fabric[\\|\/].*|(JVM|Dockerized)Service\.java$)"/>
```
derklaro, line 23: this is needed as checkstyle doesn't understand statements before super calls yet

node/src/main/java/eu/cloudnetservice/node/service/defaults/AbstractService.java
```diff
@@ -782,6 +785,33 @@ protected void downloadInclusionFile(@NonNull ServiceRemoteInclusion inclusion,
       this.serviceId().nodeUniqueId()};
   }
 
+  protected void initStandardServiceLogHandler() {
+    this.logCache.addHandler((_, line, stderr) -> {
+      for (var logTarget : this.logTargets) {
+        if (logTarget.first().equals(ChannelMessageSender.self().toTarget())) {
```
0utplay, line 791: I know that this was just moved, but can we create the ChannelMessage target once instead of each time?

0utplay (approved):

PR #1508, opened 2024-09-18
fix query packets sent from wrapper with targetAll

DasBabyPixel:
### Motivation

Fixes an issue I encountered where you can't respond to a query packet on the node if it is sent from the wrapper with `targetAll()`, because the node's response content buffer is cleared to early.

### Modification

Clear the buffer after it has been sent back to the wrapper, not before. Also adds logging, because any errors in the `CompletableFuture` callback won't be logged, making it difficult to see problems

### DIFF
--- a/node/src/main/java/eu/cloudnetservice/node/network/listener/PacketServerChannelMessageListener.java
+++ b/node/src/main/java/eu/cloudnetservice/node/network/listener/PacketServerChannelMessageListener.java
@@ -31,9 +31,13 @@ import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import lombok.NonNull;
 import org.jetbrains.annotations.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Singleton
 public final class PacketServerChannelMessageListener implements PacketListener {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(PacketServerChannelMessageListener.class);
 
   private final NodeMessenger messenger;
   private final EventManager eventManager;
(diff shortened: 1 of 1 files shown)


derklaro (changes requested):
imo it would be better to move the logging and releasing of the initial response message into a subsequent whenComplete step. This also prevents a memory leak in case an exception was caught earlier in the processing steps and therefore the code for releasing the content is never called. Also the comment when releasing the initial response content in case it was not sent to a channel is wrong now, i think it can be removed now as it is pretty clear whats happening

node/src/main/java/eu/cloudnetservice/node/network/listener/PacketServerChannelMessageListener.java
```diff
@@ -119,18 +123,27 @@ private void resumeHandling(
           } else {
             // just respond with nothing when an exception was thrown
             responseContent = DataBuf.empty().writeBoolean(false);
+            LOGGER.error("An error occurred while sending a query message", exception);
```
derklaro:
```suggestion
            LOGGER.error("Unable to relay channel message {} into cluster", message, exception);
```

node/src/main/java/eu/cloudnetservice/node/network/listener/PacketServerChannelMessageListener.java
```diff
@@ -119,18 +123,27 @@ private void resumeHandling(
...
           }
 
           // send the results to the sender
           channel.sendPacket(packet.constructResponse(responseContent));
+          return null;
+        }).whenComplete((_, exception) -> {
+          // log any internal errors
+          if (exception != null) {
+            LOGGER.error("Query response packet failed", exception);
```
derklaro:
```suggestion
            LOGGER.error("Unable to encode/send response to channel message {}", message, exception);
```

derklaro (changes requested):
Aside from the error messages this should be fine

derklaro (approved):
LGTM, thanks

PR #1521, opened 2024-10-02
non blocking output reader

DasBabyPixel:
### Motivation

Logging had some issues with blocking when a full line wasn't written in output.

### Modification

Fixed those issues by implementing a custom readLine() which calls ready() before every read()

### Result

Logging works with the new system (and very little delay, which is very nice, as opposed to 1 second polling alternative)

##### Other context

@0utplay or @GiantTreeLP please review this derklaro is fed up with this problem
Still based on the old logging to better highlight modifications. This is the reason why this is a conflict
As soon as this is reviewed I can rebase/merge and revert the revered new logging system to make this mergeable

### DIFF
--- a/modules/dockerized-services/impl/src/main/java/eu/cloudnetservice/modules/docker/impl/DockerizedService.java
+++ b/modules/dockerized-services/impl/src/main/java/eu/cloudnetservice/modules/docker/impl/DockerizedService.java
@@ -116,7 +116,8 @@ public class DockerizedService extends JVMService {
       configuration,
       manager,
       eventManager,
-      logCache, versionProvider,
+      logCache,
+      versionProvider,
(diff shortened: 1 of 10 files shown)


0utplay:
Does this catch jvm crashes?

DasBabyPixel:
> Does this catch jvm crashes?

Catches invalid jvm arguments
image

Also catches everything in a shutdown hook
image
(Used FileDescriptor, because logging frameworks don't print everything to console, probably some issue with async logging on shutdown)

DasBabyPixel:
May have some issues after considerable changes, I'm going to use this in my build first and if I think it is stable, I'm going to mark this ready for review.
The reason why the old logging reader stinks: it doesn't catch jvm crashes.

PR #1522, opened 2024-10-04
fix: generated velocity meta file uses name as version

byPixelTV:
### Motivation
Fixing an issue where the version of a Velocity plugin whould be the name set in the @PlatformPlugin annotation

### Modification
Changed the eu/cloudnetservice/ext/platforminject/processor/platform/velocity/VelocityPluginInfoGenerator.java file to now set the version to the version and not to the name

### Result
It now should set the version correctly


derklaro (approved):

derklaro:
nice catch, thanks!

PR #1523, opened 2024-10-07
fix: signs checked configuration templates instead of installed templates

Slxvly:
### Motivation
The 3.5 documentation for signs (https://cloudnetservice.eu/de/docs/next/modules/signs) says: "The tags can even be targeted to specific groups with a template path limitation, which will then only show services with the specified template." and I think that should still be valid for CloudNet 4.0. The problem is that the current implementation checks for templates in the configuration, which doesn't necessarily match the installed templates (e.x. using the smart module template configuration can change that).

### Modification
Changed the getter for templates to only get installed templates in the templatePath check.

### Result
Now, only templates that are actually installed will make a service show up on signs with the specified templatePath. This is also consistent with the templates that are displayed when the service command is used to look up its information.


0utplay:
Hi,
thanks for your pull request. While the solution you provided is correct it has some impact. You might have noticed that the installedTemplates are only available using the provider of a service. This means that each of these installedTemplates calls is a RPC request to the node of the service calling it and maybe also even to other nodes if the service is not running on the same node.

We should be really careful with these requests on the wrapper. We will take a look if we might be able to solve this another way

PR #1540, opened 2024-10-22
fix: simple v4 docker image build 

NobleMajo:
# fix: simple v4 docker image build 
### Motivation
Using and debugging nightly in a Docker container environment.
I am getting build errors due to missing build tools and I think the jre is not enough to build cloudnet.

### Modification
- using jdk for the docker multipart build image
- remove legacy docker compose version

### Result
After using the jdk to build it, it works, what a miracle!

##### Other context
No, **a simple fix**


0utplay:
Using a CRAC image is probably not the right tool for the job.

PR #1573, opened 2025-01-20
fix: remove session if chunked file transfer cannot be initiated

DasBabyPixel:
### Motivation
When a chunked file transfer is started a session is registered into the local session registry. However, if no data can be provided from the remote peer (e.g. if a template doesn't exist) the session is not unregistered from the local session registry.

### Modification
Unregister a session from the local session registry if a chunked file transfer cannot be started.

### Result
No dead chunked file transfer sessions in local session registry.


driver/impl/src/main/java/eu/cloudnetservice/driver/impl/network/chunk/builder/DefaultChunkedFileQueryBuilder.java
```diff
@@ -146,6 +146,8 @@ public class DefaultChunkedFileQueryBuilder implements ChunkedFileQueryBuilder {
           // transfer started successfully
           return responseFuture;
         } else {
+          // remove session from registry if transfer failed. Usually because the file/template is not found
+          sessionRegistry.completeSession(sessionId);
           // transfer couldn't be started for some reason
```
derklaro:
```suggestion
          // transfer couldn't be started for some reason
          sessionRegistry.completeSession(sessionId);
```

derklaro (approved):

PR #1575, opened 2025-01-20
Fix/template storage rpc

DasBabyPixel:
### Motivation

RemoteTemplateStorage doesn't work with async transfers after refactor

### Modification

Fix RemoteTemplateStorage
Serialize `java.nio.file.Path` as absolute path in transfers. Relative paths make no sense, the working directory changes when transferring

### Result

Everything works again as it should


derklaro:
superseded by #1582

PR #1582, opened 2025-01-26
fix: properly override template storage implementations for remote

0utplay:
### Motivation
When api-impl splitting the RemoteTemplateStorage impl was not done correctly. Methods that handle file streams (in & out) were just called as usual RPCs which is not possible.

### Modification
This PR overrides the specific method implementations and uses the chunked file transfer api to transfer filestreams.

### Result
The template storage is working properly from the wrapper side.

### DIFF
--- a/driver/impl/src/main/java/eu/cloudnetservice/driver/impl/template/RemoteTemplateStorage.java
+++ b/driver/impl/src/main/java/eu/cloudnetservice/driver/impl/template/RemoteTemplateStorage.java
@@ -207,11 +207,8 @@ public abstract class RemoteTemplateStorage implements TemplateStorage {
     @NonNull Path directory,
     @Nullable Predicate<Path> filter
   ) {
-    try (var inputStream = ZipUtil.zipToStream(directory, filter)) {
-      return this.deployAsync(target, inputStream);
-    } catch (IOException exception) {
-      return CompletableFuture.completedFuture(false);
-    }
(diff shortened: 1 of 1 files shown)


driver/impl/src/main/java/eu/cloudnetservice/driver/impl/template/RemoteTemplateStorage.java
```diff
@@ -140,22 +170,63 @@ public boolean deploy(@NonNull ServiceTemplate target, @NonNull InputStream inpu
...
+  @Override
+  public @NonNull CompletableFuture<Boolean> deployDirectoryAsync(
+    @NonNull ServiceTemplate target,
+    @NonNull Path directory,
+    @Nullable Predicate<Path> filter
+  ) {
+    try (var inputStream = ZipUtil.zipToStream(directory, filter)) {
+      return this.deployAsync(target, inputStream);
+    } catch (IOException exception) {
+      return CompletableFuture.completedFuture(false);
```
derklaro: Shouldn't this return a failed future with the exception?

derklaro: I guess the zipToStream should be in a supplyAsync call that composes the deploy step afterwards? something like:
```java
return CompletableFuture.supplyAsync(() -> ZipUtil.zipToStream(directory, filter))
  .thenCompose(inputStream -> this.deployAsync(target, inputStream));
```

derklaro (approved):

PR #1605, opened 2025-04-03
improve: service console output reading

derklaro:
### Motivation
Currently service process output streams are read once per second on a single shared thread. This causes issues with delayed logs, especially when the JVM crashes at startup as 99% of the times the JVM crash reason wasn't read from the process error stream. This leaves users with nearly undiagnosable startup issues.

Example log output with an invalid JVM flag before the change:
```
4.0.0-RC11-SNAPSHOT-f66323ca => create by Proxy 1 --start
[26.08 09:26:50.417] INFO : Starting to create 1 services for Proxy
[26.08 09:26:50.641] INFO : CloudService [uniqueId=04789662-d056-48ba-975a-f69993ed56dd task=Proxy name=Proxy-1] has been started
[26.08 09:26:50.642] INFO : The services were created based on the task. They can be managed with the service command
[26.08 09:26:50.703] INFO : CloudService [uniqueId=04789662-d056-48ba-975a-f69993ed56dd task=Proxy name=Proxy-1] has been stopped
```

### Modification
Now the log cache uses a scaling task executor paired with a custom non-blocking input stream reader to read the console output produced by services. Due to the better buffering and faster read times, more log output should be catched more quickly which also includes jvm crashes. The result is not perfect, but way better than it was before.

Additionally, empty log lines are now ignored in the parent log cache class and there is no need anymore for log caches to implement the handling themselfes. Also a small issue with caching was fixed that caused the cache to still contain 1 item even if the log cache entry limit was set to 0.

### Result
Log output of services is not read immediately once data is available from the input or output stream, making the output more responsive (as it is nearly immediately printed) and also catches everything that is being logged.

Example log output with an invalid JVM flag & this change applied:
```
4.0.0-RC11-SNAPSHOT-b99d7fbc => create by Proxy 1 --start
[26.08 09:21:59.245] INFO : Starting to create 1 services for Proxy
[26.08 09:21:59.289] INFO : CloudService [uniqueId=3a718134-d248-41d4-9f40-75751e1c1d64 task=Proxy name=Proxy-1] has been started
[26.08 09:21:59.289] INFO : The services were created based on the task. They can be managed with the service command
[26.08 09:21:59.295] WARN : [Proxy-1/WARN]: Unrecognized option: -XIReallyKnowWhatIAmDoingISwear
[26.08 09:21:59.296] WARN : [Proxy-1/WARN]: Error: Could not create the Java Virtual Machine.
[26.08 09:21:59.296] WARN : [Proxy-1/WARN]: Error: A fatal exception has occurred. Program will exit.
[26.08 09:22:00.303] INFO : CloudService [uniqueId=3a718134-d248-41d4-9f40-75751e1c1d64 task=Proxy name=Proxy-1] has been stopped
```


node/impl/src/main/java/eu/cloudnetservice/node/impl/service/defaults/log/NonBlockingLineReader.java
```diff
@@ -0,0 +1,188 @@
...
+    // if the buffer is closed, try to read a line
+    // from the remaining buffered content
+    var reader = this.reader;
+    if (reader == null) {
+      return this.readLineFromBuffer();
+    }
+
+    // if the input is ready read all chars that are available right away
+    if (reader.ready()) {
+      var charsRead = reader.read(this.charBuffer);
```
DasBabyPixel, line 74: pretty sure this right here is an issue, gonna confirm this later when I'm back home. Just wanted to say that before it gets merged

DasBabyPixel:
```
public static void main(String[] args) throws IOException {
    var stream = new InputStream() {
        private int idx = 0;
        private int[] b = {'A', 'B', 'C', 'D', 'E'};

        @Override
        public int read() throws IOException {
            while (idx >= b.length) {
                LockSupport.park();
            }
            return b[idx++];
        }

        @Override
        public int available() throws IOException {
            return b.length - idx;
        }
    };
    var reader = new InputStreamReader(stream);

    var cbuf = CharBuffer.allocate(1024);
    System.out.println("Start read");
    System.out.println("Ready: " + reader.ready());
    var read = reader.read(cbuf);
    System.out.println("Read: " + read);
    System.out.println(cbuf);
}
```
This should simulate what this PR does, but this blocks. Maybe my understanding of your implementation is wrong, correct me if that is the case.

Efficient reading can be implemented by using a [`InputStream#available`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/InputStream.html#available()).

> Returns an estimate of the number of bytes that can be read (or skipped over) from this input stream without blocking, which may be 0, or 0 when end of stream is detected. The read might be on the same thread or another thread. A single read or skip of this many bytes will not block, but may read or skip fewer bytes. 

Probably gonna have to give up on using a Reader in that case. Shouldn't be too much of an issue, the reader has no real use except decoding bytes to strings anyway, since we are already using a `CharBuffer` here. So the reader just ends up wrapping the InputStream, and hiding functionality / potentially caching some bytes, which may be unwanted.
Decoding bytes to strings can be done manually 🤷

DasBabyPixel: Turns out I'm wrong here, this is non-blocking

0utplay (approved):

PR #1611, opened 2025-04-17
Update gradle build to support new gradle features + use configuration cache

DasBabyPixel:
### Motivation

I wanna be able to build this on my laptop without waiting for us to collide with the andromeda galaxy.
Also the build task did not properly build everything. You need to call `build shadowJar` atm.

### Modification

 - Use gradle 9.1.0
 - Use gradle daemon JVMs
 - Removed many overlapping task outputs by specifying the exact task instead of using `withType<Jar>()`. The overlaps were mainly caused by shadowJar and jar, which always makes them not up-to-date, slowing down the build.
 - Enabled configuration cache
 - Removed all usages of `subprojects` and `allprojects`. The equivalent is now in `build-extensions` using the gradle lifecycle allprojects and then filtering for project paths.
 - Added develocity plugin + automatic build scans (only if file gradle.tos.agree exists with ToS URL as content)
 - Enabled [Configuration on Demand](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html)
 - Started to move towards support for [Isolated Projects](https://docs.gradle.org/current/userguide/isolated_projects.html) be removing usage of shared mutable state, and instead using a supported alternative, like consumable configurations
 - Fixed task execution, `build` now actually builds everything. To be more precise: `assemble` builds everything, `build` is `assemble` + all tests/checks
 - A lot of decoupling. This should help to increase the amount of work gradle can do parallel, decreasing build time.
 - Moved `exportCnlFile` and `exportLanguageFileInformation` into tasks
 - Added coroutines for exportCnlFile. That task was taking >20s because it was resolving repositories in series. Not anymore...
 - use `.configureEach` everywhere to make task configuration lazy
 
### Remaining work
 - Decouple more things, so configure on demand actually helps (I think spotless also ruins configure on demand, to be investigated). This is especially helpful when only compiling the node, as the modules are completely ignored and do not slow down the build.

### Result

Overall better build performance and no more deprecated features used.


derklaro:
while we're at it anyway, can we ignore the scaffold projects from deployment too? e.g. for the wrapper, only the wrapper-api and wrapper-impl projects should be published, but not the "glue" wrapper module.

DasBabyPixel:
Yes, I can change that! Still waiting on indra... Unless we create/publish a fork just for that 1 line...

DasBabyPixel:
Correction: Apparantly indra v4 should already be configuration cache compatible.
I'll try later to move to indra v4 published to
https://repo.stellardrift.ca/maven/#/snapshots/net/kyori/indra-git/4.0.0-SNAPSHOT
https://github.com/KyoriPowered/indra/blob/c43317c8f5f5d73e50cce414b2be3c6a6c8100e7/build.gradle#L98C41-L98C91

DasBabyPixel:
> Correction: Apparantly indra v4 should already be configuration cache compatible. I'll try later to move to indra v4 published to https://repo.stellardrift.ca/maven/#/snapshots/net/kyori/indra-git/4.0.0-SNAPSHOT https://github.com/KyoriPowered/indra/blob/c43317c8f5f5d73e50cce414b2be3c6a6c8100e7/build.gradle#L98C41-L98C91

I lied btw don't know what I thought there when I was already using a modified 4.0.0-SNAPSHOT because the normal one didn't work 🤔

derklaro:
@DasBabyPixel thanks for working on this! 🙂

PR #1616, opened 2025-05-05
fix: properly read docker task config from service props & task props

0utplay:
### Motivation
In #1251 an issue regarding docker configs was resolved. The docker configuration was only read from the service props which is not used by default. The properties are only present in the task and not the service. This PR now fixes the issue and brings support to resolve the config from service & task properties.

### Modification
Added a three step process to figure out which configuration to use when resolving properties that might be task / service specific.

The order is now:
Service -> Task (if present) -> Module Config

### Result
The configuration is read correctly


0utplay:
This PR works in a way that if the config is present in the service or the task, then all values are taken from that config. No merging of e.g. exposed ports is done. This might need to be discussed

derklaro (approved):

PR #1643, opened 2025-05-29
Make the useful of suggestion in node console from services.

Score2:
### Motivation
As the title says, it enables node console to directly get suggestions from the service, especially when you need to use cloudnet to debug the service.

The java sub-process started by cloudnet does not have a complete terminal environment, so it cannot directly get the service suggestions. Therefore, it can only communicate with the bridge through `ChannelMessage` to get suggestions.

> btw, I think service suggestions are more useful on the web than on the node console, like https://github.com/docimin/cloudnet-webinterface

### Modification

* Added suggestion processor `ServiceCommand#suggestCommands`
* Added a method `SpecificCloudServiceProvider#consoleSuggestion` in wrapper, Implementation in `AbstractService` to request suggestions from `bridge`.
* Fixed suggestion-processor for `@Greedy` processes in `DefaultSuggestionProcessor`

- [x] Bukkit
- [x] BungeeCord
- [x] Velocity
- [ ] Sponge
- [ ] Fabric
- [ ] Limbo(LOOPHP)
- [ ] Nukkit
- [ ] WaterDog
- [ ] Minestom

### Result
Perfect. Like these:

| Bukkit | BungeeCord | Velocity |
|-------|-------|-------|
| screenshot_ 2025-05-30 at 01 50 17 | screenshot_ 2025-05-30 at 01 50 44 | screenshot_ 2025-05-30 at 01 51 20 |

> [demo video like this](https://cdn.discordapp.com/attachments/818777626663321671/1377559779899215872/Untitled.mov?ex=68396805&is=68381685&hm=a5b7210baa06d3733f1abe7d4ee6eb155f80615c9cf698d50fcf9a94f94a1cc1&)

##### Other context
Once merged, [module-rest](https://github.com/CloudNetService/module-rest) will be directly compatible with this change, further enhancing the [web-interface](https://github.com/docimin/cloudnet-webinterface) console of service.

### DIFF
--- a/node/impl/src/main/java/eu/cloudnetservice/node/impl/command/defaults/DefaultSuggestionProcessor.java
+++ b/node/impl/src/main/java/eu/cloudnetservice/node/impl/command/defaults/DefaultSuggestionProcessor.java
@@ -54,8 +54,11 @@ final class DefaultSuggestionProcessor implements SuggestionProcessor<CommandSou
     @NonNull CommandPreprocessingContext<CommandSource> context,
     @NonNull Stream<Suggestion> allSuggestions
   ) {
+    var commandInput = context.commandInput();
+    // fix the suggestion of greedy
+    commandInput.cursor(commandInput.input().length());
     // íf there is no input yet, just return all suggestions
-    var input = context.commandInput().peekString();
+    var input = commandInput.peekString();
     if (Strings.isNullOrEmpty(input)) {
       return allSuggestions;
     }
--- a/node/impl/src/main/java/eu/cloudnetservice/node/impl/command/defaults/DefaultSuggestionProcessor.java
+++ b/node/impl/src/main/java/eu/cloudnetservice/node/impl/command/defaults/DefaultSuggestionProcessor.java
@@ -55,8 +55,13 @@ final class DefaultSuggestionProcessor implements SuggestionProcessor<CommandSou
     @NonNull Stream<Suggestion> allSuggestions
   ) {
     var commandInput = context.commandInput();
+
     // fix the suggestion of greedy
-    commandInput.cursor(commandInput.input().length());
+    var lastWhitespace = commandInput.input().lastIndexOf(' ');
+    if (lastWhitespace != -1) {
+      commandInput.cursor(lastWhitespace);
+    }
+
     // íf there is no input yet, just return all suggestions
     var input = commandInput.peekString();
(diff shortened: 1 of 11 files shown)


node/impl/src/main/java/eu/cloudnetservice/node/impl/command/defaults/DefaultSuggestionProcessor.java
```diff
@@ -54,8 +54,16 @@ final class DefaultSuggestionProcessor implements SuggestionProcessor<CommandSou
     @NonNull CommandPreprocessingContext<CommandSource> context,
     @NonNull Stream<Suggestion> allSuggestions
   ) {
+    var commandInput = context.commandInput();
+
+    // fix the suggestion of greedy
+    var lastWhitespace = commandInput.input().lastIndexOf(' ');
+    if (lastWhitespace != -1) {
+      commandInput.cursor(lastWhitespace);
```
Score2, line 62: It's worth mentioning that this is a potentially destructive fix.

actually, I've tried setting the `cursor` in the `Parser`, but it obviously doesn't work for the `Greedy` argument, because [CommandTree#L841(Incendo/cloud)](https://github.com/Incendo/cloud/blob/3c83f9a680609299894406a2de2502209c16f113/cloud-core/src/main/java/org/incendo/cloud/CommandTree.java#L841) restores the `cursor` modified in the `Parser`.

at least in my opinion, there is no better way to solve it properly.

node/impl/src/main/java/eu/cloudnetservice/node/impl/service/defaults/AbstractService.java
```diff
@@ -411,6 +412,19 @@ public void restart() {
     this.updateLifecycle(ServiceLifeCycle.RUNNING);
   }
 
+  @Override
+  public Collection<String> consoleSuggestion(@NonNull String line) {
+    var listPropertyType = TypeFactory.parameterizedClass(List.class, String.class);
```
derklaro, line 417: I think this method should return a future instead so that the caller can better handle errors and apply timeouts.

derklaro (commented):
The biggest issue I have with the current implementation is that this introduces a dependency of the node on the bridge, which might not be installed. I think it would be better if the node provides an interface that can be implemented by modules and resolves it using the service registry. This way a module can hook into the node rather than the node hooking into the bridge.

PR #1647, opened 2025-06-02
fix: logging configuration properties not working from launcher.cnl

ttugrul:
Motivation
Logging properties set in launcher.cnl (like var cloudnet.log.level OFF) were being ignored.

Modification
Changed logback XML files to read system properties directly instead of using cached variables:
${cloudnet.log.level} → ${cloudnet.log.level:-DEBUG}
${cloudnet.log.path} → ${cloudnet.log.path:-local/logs}

Result
Logging configuration from launcher.cnl now works correctly. Users can set custom log levels and paths.

Other context
Fixes timing issue where logback initialized before system properties were available.

### DIFF
--- a/node/impl/src/main/resources/logback.xml
+++ b/node/impl/src/main/resources/logback.xml
@@ -22,10 +22,6 @@
   <import class="eu.cloudnetservice.node.impl.console.log.ConsoleLogAppender"/>
   <import class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy"/>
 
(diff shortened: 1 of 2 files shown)


derklaro:
I wonder if it would be better to move the variable subsitutation into the variable values, so that they are defined at the top of the configuration. I think the same result can be achived using something like:
```xml
<variable name="cloudnet.log.level" value="${cloudnet.log.level:-DEBUG}"/>
```

0utplay (approved):

derklaro:
@ttugrul thanks!

PR #1685, opened 2025-07-04
chore: update to junit v6

derklaro:
### Motivation
JUnit v6 was released and we should update to it.

### Modification
Update to junit v6.

### Result
We're using the latest junit version.


derklaro:
only open to track junit v6 milestones & the pmigration progress.

0utplay (dismissed):

PR #1687, opened 2025-07-14
chore: improve channel messaging and address memory leaks

### LINKED ISSUE #1577: Memory leaks because the packet content buffer is not released in handlers
### Stacktrace

```
[21.01 20:14:49.787] ERROR: LEAK: Object "buffer (0 bytes)" was not property closed before it was garbage collected. A life-cycle back-trace (if any) is attached as suppressed exceptions. See https://netty.io/wiki/reference-counted-objects.html for more information.
io.netty5.buffer.LoggingLeakCallback$LeakReport: Object life-cycle trace:
        Suppressed: io.netty5.buffer.internal.LifecycleTracer$Traceback: ALLOCATE (current acquires = 0) T-60132541us.
                at io.netty5.buffer.internal.ResourceSupport.<init>(ResourceSupport.java:42)
                at io.netty5.buffer.internal.AdaptableBuffer.<init>(AdaptableBuffer.java:28)
                at io.netty5.buffer.bytebuffer.NioBuffer.<init>(NioBuffer.java:65)
                at io.netty5.buffer.bytebuffer.ByteBufferMemoryManager.createBuffer(ByteBufferMemoryManager.java:85)
                at io.netty5.buffer.bytebuffer.ByteBufferMemoryManager.allocateShared(ByteBufferMemoryManager.java:59)
                at eu.cloudnetservice.driver.impl.network.netty.buffer.NettyNioBufferReleasingAllocator.allocate(NettyNioBufferReleasingAllocator.java:83)
                at io.netty5.buffer.bytebuffer.NioBuffer.copy(NioBuffer.java:205)
                at io.netty5.buffer.Buffer.copy(Buffer.java:766)
                at eu.cloudnetservice.driver.impl.network.netty.codec.NettyPacketDecoder.decode(NettyPacketDecoder.java:67)
                at io.netty5.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:387)
                at io.netty5.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:330)
                at io.netty5.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:204)
                at io.netty5.channel.DefaultChannelHandlerContext.invokeChannelRead(DefaultChannelHandlerContext.java:455)
                at io.netty5.channel.DefaultChannelHandlerContext.findAndInvokeChannelRead(DefaultChannelHandlerContext.java:445)
                at io.netty5.channel.DefaultChannelHandlerContext.fireChannelRead(DefaultChannelHandlerContext.java:426)
                at io.netty5.channel.internal.DelegatingChannelHandlerContext.fireChannelRead(DelegatingChannelHandlerContext.java:113)
                at io.netty5.handler.codec.ByteToMessageDecoder$ByteToMessageDecoderContext.fireChannelRead(ByteToMessageDecoder.java:446)
                at eu.cloudnetservice.driver.impl.network.netty.codec.VarInt32FrameDecoder.decode(VarInt32FrameDecoder.java:65)
                at io.netty5.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:387)
                at io.netty5.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:330)
                at io.netty5.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:204)
                at io.netty5.channel.DefaultChannelHandlerContext.invokeChannelRead(DefaultChannelHandlerContext.java:455)
                at io.netty5.channel.DefaultChannelHandlerContext.findAndInvokeChannelRead(DefaultChannelHandlerContext.java:445)
                at io.netty5.channel.DefaultChannelHandlerContext.fireChannelRead(DefaultChannelHandlerContext.java:426)
                at io.netty5.channel.ChannelHandler.channelRead(ChannelHandler.java:235)
                at io.netty5.channel.DefaultChannelHandlerContext.invokeChannelRead(DefaultChannelHandlerContext.java:455)
                at io.netty5.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:838)
                at io.netty5.channel.AbstractChannel$ReadSink.processRead(AbstractChannel.java:1975)
                at io.netty5.channel.epoll.EpollSocketChannel.epollInReadyBytes(EpollSocketChannel.java:443)
                at io.netty5.channel.epoll.EpollSocketChannel.epollInReady(EpollSocketChannel.java:411)
                at io.netty5.channel.epoll.AbstractEpollChannel.doReadNow(AbstractEpollChannel.java:305)
                at io.netty5.channel.AbstractChannel$ReadSink.readLoop(AbstractChannel.java:2035)
                at io.netty5.channel.AbstractChannel.readNow(AbstractChannel.java:910)
                at io.netty5.channel.epoll.AbstractEpollChannel.epollInReady(AbstractEpollChannel.java:296)
                at io.netty5.channel.epoll.EpollHandler.processReady(EpollHandler.java:452)
                at io.netty5.channel.epoll.EpollHandler.run(EpollHandler.java:371)
                at io.netty5.channel.SingleThreadEventLoop.runIO(SingleThreadEventLoop.java:192)
                at io.netty5.channel.SingleThreadEventLoop.run(SingleThreadEventLoop.java:176)
                at io.netty5.util.concurrent.SingleThreadEventExecutor.lambda$doStartThread$4(SingleThreadEventExecutor.java:774)
                at io.netty5.util.internal.ThreadExecutorMap.lambda$apply$1(ThreadExecutorMap.java:68)
                at io.netty5.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
                at java.base/java.lang.Thread.run(Thread.java:1575)
```

### Actions to reproduce

Enable leak detection
```
var cloudnet.net.leak-detection-enabled true
var io.netty5.buffer.lifecycleTracingEnabled true
```

### CloudNet version

too lazy to boot up an unmodified version rn.
no, I am not imagining nor creating the leaks, that much I am sure of

### Other

Basically after decoding query packets dont get released and RPC may in some cases also not release the buffer

### Issue uniqueness

- [x] Yes, this issue is unique. There are no similar issues.

derklaro:
### Motivation
The channel messaging system (as well as a few network parts in general) really need a cleanup to improve the code quality and reliability of the system. Furthermore, there were memory leaks reported that need to be addressed as well.

### Modification
This pr mostly improves channel messaging to be more reliable overall and improves the api surface exposed to consumers. Furthermore, all possible memory leaks were addressed (not limited to channel messaging) and some other networking parts were improved as well.

### Result
Channel messaging should be more reliable and easier to use. Additionally, there shouldn't be any memory leaks anymore.


driver/impl/src/main/java/eu/cloudnetservice/driver/impl/network/chunk/network/ChunkedPacketListener.java
```diff
@@ -59,10 +59,21 @@ public void handle(@NonNull NetworkChannel channel, @NonNull Packet packet) thro
     var chunkIndex = packetContent.readInt();
 
     // get or create a new local session for the transfer
-    var sessionHandler = this.sessionRegistry.getOrCreateSession(sessionInfo, this.handlerFactory);
+    var holder = new Object() {
+      boolean factoryCalled = false;
+    };
+    var sessionId = sessionInfo.sessionUniqueId();
+    var sessionHandler = this.sessionRegistry.getOrCreateSession(sessionId, () -> {
+      holder.factoryCalled = true;
```
DasBabyPixel, line 67: whhaaat var can infer anonymous classes amd their attributes 👀

driver/impl/src/main/java/eu/cloudnetservice/driver/impl/network/netty/NettyNetworkHandler.java
```diff
@@ -76,9 +76,25 @@ public void channelReadComplete(@NonNull ChannelHandlerContext ctx) {
...
+   * registry, calling all associated handlers. This method applies exception handling which is not done by
+   * {@link #doHandlePacket(BasePacket)}.
+   *
+   * @param packet the packet hto handle.
+   * @throws NullPointerException if the given packet is null.
+   */
+  private void handlePacket(@NonNull BasePacket packet) {
+    try {
+      this.doHandlePacket(packet);
+    } catch (Exception exception) {
```
DasBabyPixel, line 96: I'd go further and catch Throwable instead of silently ignoring OOMEs or similar

driver/impl/src/main/java/eu/cloudnetservice/driver/impl/network/netty/buffer/NettyImmutableDataBuf.java
```diff
@@ -279,12 +275,10 @@ public int acquires() {
...
   public void release() {
-    // release one acquire
-    this.acquires--;
-
-    // check if the buffer is no longer acquired somewhere
-    if (this.acquires <= 0 && this.buffer.isAccessible()) {
+    try {
       this.buffer.close();
+    } catch (IllegalStateException _) {
+      // possible double-free error due to a race, ignore
```
DasBabyPixel, line 281: since when are buffers thread safe? This seems like  a hack that will cause issues later

modules/bridge/impl/src/main/java/eu/cloudnetservice/modules/bridge/impl/platform/helper/ProxyPlatformHelper.java
```diff
@@ -44,15 +44,20 @@ public ProxyPlatformHelper(@NonNull ComponentInfo componentInfo) {
...
+    var response = this.toCurrentNode()
       .message("proxy_player_pre_login")
       .channel(BridgeManagement.BRIDGE_PLAYER_CHANNEL_NAME)
       .buffer(DataBuf.empty().writeObject(playerInfo))
       .build()
       .sendSingleQuery();
-    return result == null
-      ? LocalPlayerPreLoginEvent.Result.allowed()
-      : result.content().readObject(LocalPlayerPreLoginEvent.Result.class);
+    return switch (response) {
```
DasBabyPixel, line 53: 😂

modules/npcs/impl/src/main/java/eu/cloudnetservice/modules/npc/impl/platform/PlatformNPCManagement.java
```diff
@@ -89,14 +88,22 @@ public PlatformNPCManagement(
       .targetNode(componentInfo.nodeUniqueId())
       .build()
       .sendSingleQuery();
-    return response == null ? null : response.content().readObject(NPCConfiguration.class);
+    return switch (response) {
```
DasBabyPixel, line 91: 😂

DasBabyPixel (commented):
I think this is a step in the right direction. If you give me a few days time I have a reason to update past RC11 other than java 24 and time to test this branch.
I think replacing null checks with pattern matching is either an overly aggressive inspection or a too idealistic world view. It does not make the code more readable.
I haven't yet gone through all usages to look for all possible leaks, but the obvious ones I do remember should be fixed here.

driver/impl/src/main/java/eu/cloudnetservice/driver/impl/network/netty/NettyNetworkHandler.java
DasBabyPixel: we should get oracle to have the default UncaughtExceptionHandler call System.exit.
I remember wasting way too many hours because network threads ate my errors and I couldn't for the life of me figure out where.

driver/impl/src/main/java/eu/cloudnetservice/driver/impl/network/netty/buffer/NettyImmutableDataBuf.java
derklaro: The buffer is not thread-safe, but the drop guarding the buffer is. So concurrent `close` calls are safe. However, I would also argue that this is the better approach as `close` calls should not throw an exception if the buffer was already closed before (this was partly implemented using the `isAccessible()` check before - just not thread-safe).

DasBabyPixel: My understanding of how the buffers work may be incorrect, but concurrent close calls can never be safe?

EDIT: part of internal buffer code in `ResourceSupport` class, `XXXBuffer extends ResourceSupport`
```java
    @Override
    public final void close() {
        if (acquires == -1) {
            throw attachTrace(new IllegalStateException("Double-free: Resource already closed and dropped."));
        }
        int acq = acquires;
        acquires--;
        if (acq != 0) {
            // Only record a CLOSE if we're not going to record a DROP.
            tracer.close(acq);
        } else {
            // The 'acquires' was 0, now decremented to -1, which means we need to drop.
            tracer.drop(0);
            try {
                drop.drop(impl());
                Reference.reachabilityFence(this); // Avoid racing with any Cleaner.
            } finally {
                makeInaccessible();
            }
        }
    }
```
`acquires` is a non-volatile int. So a buffer must, at every point in time, have a clear "owner" thread. Having IllegalStateExceptions thrown seems to me like the presence of another issue somewhere else, maybe in how the buffer is used or passed between owners.

DasBabyPixel: If we want to enable concurrent close calls, we need to have some sort of atomic isClosed flag or counter

derklaro: Yes, that place is not atomic, but the actual dropping happens in [ArcDrop](https://github.com/netty/netty/blob/main/buffer/src/main/java/io/netty5/buffer/internal/ArcDrop.java) (in our case) which is.

DasBabyPixel: I guess you are right, ArcDrop is atomic and concurrent close calls should be safe. 
Still think this is a bad idea long-term and while it may work now may fail if the netty implementation details change. Someone once told me relying on implementation details is a hassle and I should avoid it, so I do.
grafik
Imo a buffer lifecycle should be clearly defined and a close call must be on the owner thread, or synchronized in some way.
Also I haven't checked, but does enabling leak detection also enable it for the wrapper?

derklaro: I don't really understand the problem. As stated before the exception is caught to ensure that, at any point, a call to `close` does **NOT** throw an exception. The previous `isAccessible()` guard did the same thing but left the door open a tiny bit open for an exception to be thrown anyway - this is now not possible anymore. Also note that the documentation of DataBuf still states that "buffers are not required to be thread safe", and there is no documentation hinting that close calls are thread-safe. It's just a security harness to be 100% sure that no exception is thrown, ever.

DasBabyPixel: If we argue that buffers are not thread safe, then concurrent close calls should be able to fail and throw the exception.
It's like instead of getting a ConcurrentModificationException when concurrently modifying an array list, it just silently fails. 
The presence of such an exceptions indicates a deeper problem, in that case the concurrent usage of close. The absence of such an exception makes is so that problems, like incorrect usage of buffers, and concurrent close calls, are more difficult to diagnose. 
It's not directly a problem where something doesn't work, correct usage of buffers will work. It's that catching that exception will make the future more difficult with bugs that can't be diagnosed because the exception is caught and ignored.

My point is: I want incorrect usage, in this case concurrent usage, to throw an exception. It makes it much easier to debug problems, and often prevents them alltogether.

0utplay (approved):

DasBabyPixel:
EDIT: nvm I think I merged something incorrectly into my fork

DasBabyPixel:
Disregarding opinions on coding guidelines etc, it's overall looking good. No immediate masses of memory leaks are being thrown at me.
Just don't forget to include the information about closing ChannelMessages in the changelogs.

PR #1689, opened 2025-07-16
Serialize Path objects as absolute paths for RPC

### LINKED ISSUE #1303: TemplateStorage#pull does not work in wrapper with relative paths
### Stacktrace

_No response_

### Actions to reproduce

Try to use TemplateStorage#pull in the Wrapper with a relative path as argument

### CloudNet version

[31.08 10:46:54.357] INFO:
[31.08 10:46:54.357] INFO: CloudNet Blizzard 4.0.0-RC9 f6ca4c38
[31.08 10:46:54.358] INFO: Discord: <https://discord.cloudnetservice.eu/>
[31.08 10:46:54.358] INFO:
[31.08 10:46:54.358] INFO: ClusterId: deebb2f9-****-41cd-****-246e55129822
[31.08 10:46:54.359] INFO: NodeId: Node-1
[31.08 10:46:54.359] INFO: Head-NodeId: Node-1
[31.08 10:46:54.360] INFO: CPU usage: (P/S) .26/1.23/100%
[31.08 10:46:54.360] INFO: Node services memory allocation (U/R/M): 1536/1536/4096 MB
[31.08 10:46:54.360] INFO: Threads: 44
[31.08 10:46:54.361] INFO: Heap usage: 40/256MB
[31.08 10:46:54.361] INFO: JVM: Amazon.com Inc. 17 (OpenJDK 64-Bit Server VM 17.0.4.1+9-LTS)
[31.08 10:46:54.361] INFO: Update Repo: CloudNetService/launchermeta, Update Branch: beta
[31.08 10:46:54.362] INFO:

### Other

It will not work because the relative Path gets serialized and sent to the Node where it will be deserialized again as a relative path. Now the process workingdirectory changed so the relative path lost its meaning

### Issue uniqueness

DasBabyPixel:
### Motivation

#1303 

### Modification

Only send the path as an absolute path over RPC

### Result

Relative paths can be used with pull and work correctly

##### Other context

Fixes #1303


derklaro:
I think it's an oversight that `pull` uses RPC, imo it should use the chunked data transfer like the other methods as well.

DasBabyPixel:
> I think it's an oversight that `pull` uses RPC, imo it should use the chunked data transfer like the other methods as well.

you're probably correct. I also don't know if this works well in clusters. I just know this fixes the issue for me, even if it is not ideal

derklaro:
I can fix the pull method later. I'm not sure if the path serializer is a good idea at all, maybe we should just remove it. If we leave it, then it probably shouldn't serialize absolute paths, and leave the decision about aboslute vs relative paths to the caller.

PR #1702, opened 2025-07-22
chore: reduce heap usage when doing chunked data transfer

derklaro:
### Motivation
Currently, chunked data transfers are copying the data via the heap into the target netty buffer (and vice-versa). This is very inefficient as this means that a native memory segment is copied to the heap and then back into a native memory segment. 

### Modification
Use nio Channels in combination with ByteBuffers to directly read/write the content being transferred to/from the netty buffer, completely removing the need for (almost) any heap allocations.

### Result
Reduced heap usage when doing a chunked data transfers.

### DIFF
--- a/driver/api/src/main/java/eu/cloudnetservice/driver/network/buffer/DataBuf.java
+++ b/driver/api/src/main/java/eu/cloudnetservice/driver/network/buffer/DataBuf.java
@@ -18,10 +18,13 @@ package eu.cloudnetservice.driver.network.buffer;
 
 import eu.cloudnetservice.driver.network.object.ObjectMapper;
 import java.lang.reflect.Type;
+import java.nio.ByteBuffer;
 import java.util.UUID;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 import lombok.NonNull;
+import org.jetbrains.annotations.ApiStatus;
+import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.Nullable;
 
 /**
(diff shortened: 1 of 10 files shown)


driver/api/src/main/java/eu/cloudnetservice/driver/network/buffer/DataBuf.java
```diff
@@ -257,12 +260,60 @@ public interface DataBuf extends AutoCloseable {
...
+   * @throws IllegalArgumentException  if the given delta is negative.
+   * @throws IllegalStateException     if this buffer was released.
+   * @throws IndexOutOfBoundsException if advancing by the given delta would move beyond the end of this buffer.
+   */
+  @NonNull
+  @Contract("_ -> this")
+  DataBuf advanceReaderOffset(int delta);
+
+  /**
+   * Get a byte buffer instance that shares the memory region of this buffer. The returned buffer is marked as read-only
```
DasBabyPixel:
```suggestion
   * Get a new byte buffer instance that shares the memory region of this buffer. The returned buffer is marked as read-only
```

driver/api/src/main/java/eu/cloudnetservice/driver/network/buffer/DataBuf.java
```diff
@@ -527,6 +582,61 @@ <T> DataBuf.Mutable writeNullable(
...
+     * @throws IllegalArgumentException  if the given delta is negative.
+     * @throws IllegalStateException     if this buffer was released.
+     * @throws IndexOutOfBoundsException if advancing by the given delta would move beyond the end of this buffer.
+     */
+    @NonNull
+    @Contract("_ -> this")
+    DataBuf advanceWriterOffset(int delta);
+
+    /**
+     * Get a byte buffer instance that shares the memory region of this buffer. The returned buffer can be used to read
```
DasBabyPixel:
```suggestion
     * Get a new byte buffer instance that shares the memory region of this buffer. The returned buffer can be used to read
```

0utplay (dismissed):

0utplay (approved):

PR #1719, opened 2025-08-11
feat: support all fabric server versions

derklaro:
### Motivation
There were a lot of questions asking about support for more than the latest fabric version.

### Modification
Use the nested mods feature of the fabric loader to load version-specific mods (nested mods are only loaded if possible depending on their version constraints). In all cases where no version mod is available, a fallback listener is registered on the services to mark them as online after a small delay to allow for the server to start. This comes with the downside that no player count and player data can be provided to the api, but that should not be much of a problem.

### Result
All versions of minecraft (using the fabric mod loader) are now supported by the bridge module.


ext/platform-inject-support/api/src/main/java/eu/cloudnetservice/ext/platforminject/api/defaults/BasePlatformPluginManager.java
```diff
@@ -77,9 +77,13 @@ public void constructAndLoad(@NonNull Class<? extends PlatformEntrypoint> plugin
 
     // configure the injection layer
     var callerClass = RETAINING_STACK_WALKER.walk(stream -> stream
-      .skip(2)
-      .map(StackWalker.StackFrame::getDeclaringClass)
+      .dropWhile(stackFrame -> {
+        var className = stackFrame.getClassName();
+        var isEntrypointClass = className.contains(".Generated") && className.endsWith("Entrypoint");
+        return !isEntrypointClass;
+      })
```
derklaro, line 84: drive-by improvement: due to the new extra reflection call when loading platform plugins, this method now needs to skip 3 frames. To make it resiliant against future changes, I changed it to actually searching the entrypoint class instead.

ext/platform-inject-support/api/src/main/java/eu/cloudnetservice/ext/platforminject/loader/PlatformInjectSupportLoader.java
derklaro, line 1: This change is needed due to the shadow update. It doesn't seem to be the case anymore that classes override each other in the way they did before, therefore, I just added this extra step that just uses reflection to call the runtime methods instead. Should not make a performance difference really.

gradle/libs.versions.toml
```diff
@@ -1,10 +1,9 @@
 [versions]
 
 # plugins
-shadow = "8.3.8"
+shadow = "9.0.1"
```
derklaro, line 4: Update is needed as v9 is actually capable of copying jar files into the final fat jar instead of unpacking them.

modules/bridge/impl/src/main/java/eu/cloudnetservice/modules/bridge/impl/platform/fabric/FabricBridgeInitializer.java
```diff
@@ -16,37 +16,60 @@
...
 
 @Singleton
 @PlatformPlugin(
   platform = "fabric",
   name = "CloudNet-Bridge",
   version = "@version@",
   dependencies = {
-    @Dependency(name = "fabricloader", version = ">=0.16.6"),
-    @Dependency(name = "minecraft", version = "~1.21.6"),
+    @Dependency(name = "fabricloader", version = ">=0.17.0"),
```
derklaro, line 37: v0.17 is needed due to our fallback version getter (`FabricLoader.getInstance().getRawGameVersion()`) which was introduced in that version.

plugins/luckperms/src/main/java/eu/cloudnetservice/plugins/luckperms/FabricLuckPermsPlugin.java
```diff
@@ -31,7 +31,7 @@
   version = "@version@",
   dependencies = {
     @Dependency(name = "fabricloader", version = ">=0.16.6"),
```
derklaro, line 33: I guess we can technically drop the requirement way down (to around `v0.12` iirc)

build.gradle.kts
```diff
@@ -20,7 +20,7 @@ plugins {
   id("cloudnet.parent-build-logic")
   alias(libs.plugins.spotless)
   alias(libs.plugins.nexusPublish)
-  alias(libs.plugins.fabricLoom) apply false
+  alias(libs.plugins.shadow) apply false // must be here to enforce the bundled asm version
```
derklaro, line 23: Somehow a way older asm version leaks into the classpath if we don't do this (previously loom provided the newer one) which causes the shadow remap operations to fail (relocation, minimizing, etc.)

0utplay (approved):

Issue #1354, opened 2024-02-21
Template Prototypes


HugKitten:
# Describe the new feature
### Blueprints  for templates!
I created this suggestion after finding the configuration options for templates too stiff and lacking of customization. I didn't like that I had to search for a module for each use case, and found many of the unofficial modules undocumented, unmaintained or broken. In the end, I thought it would be better for a single unified setup that can be used across any setup regardless of plugin. This means something that would work for any permission plugin, moderation plugin, database plugin, etc. I think the best way to do this is by allowing a special extension type for templates that allows for data to be used. I have documented some examples below.

### Using the task name from config files
_LuckPerms/config.yml.prototype_
```
# @name becomes the name of the task
server: @name
sync-minutes: 10
watch-files: false
```

### Using the task and group properties from config files
_LiteBans/config.yml.prototype_
```
server_name: '@properties.litebans.serverName'
default_server_scope: '@properties.litebans.defaultServerScope'
```

### Reusing velocity forwarding secret key across all instances
_forwarding.secret_
```
@properties.forwardingSecret
```

### Configuring seperate databases for test nodes and deployment nodes:
**Test config**
```
"properties": {
    "mysql": {
      "host" : "test-host",
      "port" : "test-port",
      "user" : "test-user",
      "password" : "test-password"
    }
  }
```
**Deployment config**
```
"properties": {
    "mysql": {
      "host" : "depolyment-host",
      "port" : "depolyment-port",
      "user" : "depolyment-user",
      "password" : "depolyment-password"
    }
  }
```
_TheNewEconomy/data.yml.prototype_
```
Data:
  Sync:
    Redis:
      Host: "@properties.mysql.host"
      Port: @properties.mysql.port
      User: "@properties.mysql.user"
      Password: "@properties.mysql.password"
```

# Considerations

### How should we escape the @ character in templates?
We could use an escape sequence such as @@ or \@.

**Benefits of @@**
The escape sequence is isolated to a single character.
The user won't have to use \\ for \ characters.

**Benefits of \@**
The escape sequence becomes more readable.
The escape of any character is supported.
The escape sequence allows for future modifiers such as !, #, or $ in the future.
Variable key starting with @@ are still possible.

### How do we handle files that need the .prototype extension?
There are two ways we can counter this. We could either ask the user to rename the file into .prototype.prototype, or we can use the @<prototype:version> declaration to signal that a file is a prototype.

**Benefits of .prototype.prototype hack**
[...]

PR #1814, opened 2025-12-07
Module: Replacer

whereareiam:
### Motivation

Some users expressed the need for a centralized and flexible way to replace placeholders across multiple configuration files. Until now, this often required duplicating sensitive or shared values (e.g. database credentials, service endpoints, secrets) across many different configs, making maintenance error-prone and inconvenient.

To solve this, I created the **Replacer module**, which allows defining placeholder replacement rules in a single configuration and applying them dynamically across multiple files. This eliminates the need to manually update the same values in many places and reduces the risk of configuration inconsistencies.

#1354 

### Modification

- Added a new **Replacer module** that:
  - Processes configurable replacement **rules**
  - Supports multiple **targets** (task, service, environment, group)
  - Allows defining **file patterns** for mass replacement
  - Supports multiple **placeholders per rule**
  - Allows different **search and replace strategies** (`ALL`, `FIRST`, etc.)
  - Supports **dynamic values** with nested placeholders (e.g. `%taskName%`, `%nodeId%`, `%serviceHost%`)
- Added example configuration demonstrating:
  - Lobby MOTD replacement
  - Endpoint generation
  - Forwarding secrets for Paper & Velocity
  - Multi-target shared secrets

### Result

- **Before:**  
  Placeholder values and secrets had to be duplicated manually across many different configuration files, making updates tedious and risky.

- **After:**  
  All placeholder replacements can now be managed from **a single centralized configuration**, automatically applied to all matching files and targets.

- **Impact:**
  - Much easier configuration management
  - No need to store sensitive data in multiple configs
  - Reduced risk of mismatched or outdated values

image

### DIFF
--- a/build-extensions/src/main/kotlin/eu/cloudnetservice/cloudnet/gradle/util/Files.kt
+++ b/build-extensions/src/main/kotlin/eu/cloudnetservice/cloudnet/gradle/util/Files.kt
@@ -44,6 +44,7 @@ object Files {
   const val storageSftp = "cloudnet-sftp.jar"
   const val storageS3 = "cloudnet-s3.jar"
   const val influx = "cloudnet-influx.jar"
+  const val replacer = "cloudnet-replacer.jar"
   const val node = "cloudnet.jar"
   const val nodeCnl = "cloudnet.cnl"
 }
--- a/ext/platform-inject-support/runtime/build.gradle.kts
+++ b/ext/platform-inject-support/runtime/build.gradle.kts
@@ -23,6 +23,7 @@ plugins {
 
 repositories {
   maven("https://repo.waterdog.dev/releases/")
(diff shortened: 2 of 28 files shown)


0utplay:
Hi, thank you for your contribution. I think it makes more sense if you first add this module as a third-party module in CloudNet, as we already do with the CloudNet-REST module, since we currently do not have the capacity to maintain and further develop the module if necessary.

Therefore, I would be very grateful if you could publish the module yourself and add it here https://github.com/CloudNetService/CloudNet/blob/nightly/modules.json with a PR. If you need any help with this, please let me know.

whereareiam:
This PR was closed and the module was moved to a dedicated repository.  
A follow-up PR was opened to add it as a third-party module.

https://github.com/whereareiam/Templify  
https://github.com/CloudNetService/CloudNet/pull/1824

PR #1830, opened 2025-12-25
fix: skip OS-level port check for addresses outside network namespace

ttugrul:
### Motivation
When CloudNet runs inside a containerized environment (e.g., Docker with bridge networking) with a host address that doesn't exist in the container's network namespace, the port allocation fails with "No free port found" error. This happens because `NetworkUtil.isInUse()` tries to bind a `ServerSocket` to an address that doesn't exist locally, causing all port checks to fail.

### Modification
- Added `NetworkUtil.isBindableAddress()` method to check if an address exists in the current network namespace
- Added `isPortInUseAtOsLevel()` hook in `BaseLocalCloudServiceFactory`
- Override the hook in `DockerizedLocalCloudServiceFactory` to skip OS-level port check for non-bindable addresses
- JVM services retain full port validation (unchanged behavior)

### Result
CloudNet can now correctly allocate ports when running in containerized environments with external host addresses. The service registry is used to prevent port conflicts for addresses outside the local network namespace.

### DIFF
--- a/modules/dockerized-services/impl/src/main/java/eu/cloudnetservice/modules/docker/impl/DockerizedLocalCloudServiceFactory.java
+++ b/modules/dockerized-services/impl/src/main/java/eu/cloudnetservice/modules/docker/impl/DockerizedLocalCloudServiceFactory.java
@@ -27,6 +27,7 @@ import eu.cloudnetservice.node.config.Configuration;
 import eu.cloudnetservice.node.impl.service.InternalCloudServiceManager;
 import eu.cloudnetservice.node.impl.service.defaults.factory.BaseLocalCloudServiceFactory;
 import eu.cloudnetservice.node.impl.tick.DefaultTickLoop;
+import eu.cloudnetservice.node.impl.util.NetworkUtil;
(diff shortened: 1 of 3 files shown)


derklaro:
I wonder if overriding the method in `DockerizedLocalCloudServiceFactory` would be a better option. Right now providing an address that the service cannot bind to for jvm services (running on the host) would pass the port check, start the service and then fail when binding, or am I missing something?

EDIT: Plus, the check also fails if another service also runs on the host/port combination (first check of the method), that seems undesireable for dockerized services as well.

ttugrul:
Thanks for the review! I've refactored to override in `DockerizedLocalCloudServiceFactory` as you suggested.

Added an `isPortInUseAtOsLevel()` hook in the base class that the Docker factory overrides to skip the OS check for non-bindable addresses. JVM services keep full port validation, unchanged from original behavior.

I kept the registry check because without it, multiple instances of the same task would all get allocated the same port (e.g. all get 25565), then Docker would fail on conflict. It ensures proper port incrementing.

ttugrul:
Hi, added Docker-aware port checking, queries running containers for port conflicts before allocation. Would appreciate a review when you have time.

PR #1835, opened 2026-01-08
Don't remove service files on restart

eliasfalch:
### Motivation
When autoDeleteOnStop is enabled, service files are being deleted regardless of if it's a restart or not.

### Modification
- added a new updateLifecycle method with a removeFiles boolean
- changed old updateLifecycle method to use new one

### Result
This should stop service files being deleted on a restart.


0utplay:
Hi, thanks for you pull request. Unfortunately this does not resolve the issue you were trying to solve.

The call to `doRemoveFilesAfterStop` is not whats causing services to be deleted on a restart. Its simply removing files that are expected to be removed on a stop (also for restarts) based on the task configuration. Furthermore you will never reach that condition because of the previous if statement checking for switchToDeletedOnStop. So this essentially only breaks the stop functionality when autoDeleteAfterStop is disabled.

PR #1895, opened 2026-08-13
chore: dependency updates

0utplay:


modules/bridge/impl/src/main/java/eu/cloudnetservice/modules/bridge/impl/platform/velocity/VelocityBridgePlugin.java
```diff
@@ -68,22 +68,27 @@ public VelocityBridgePlugin(
 
   @Override
   public void onLoad() {
```
0utplay, line 70: breaking with velocity 3.X I guess. Might not be worth it at this point in time

derklaro (approved):

